PR #8, opened 2018-06-09
add deterministic sampling to config, defaulting to trace_id

rfong:


rfong:
debating whether to refactor our deterministic sampling logic into libhoney-go or a separate package

toshok:
hm, honeycombio/beeline-nodejs#56 adds it to the nodejs beeline...

PR #11, opened 2018-07-25
Don't panic if the handler is a struct

apechimp:
We use structs for our handlers. Getting the function name will panic as a result. This corrects that.


nathanleclaire:
Awesome! Thank you Mike. We're going to take a look at this and will likely merge shortly.

apechimp:
You're welcome! Thanks for the prompt response.

nathanleclaire:
LGTM

nathanleclaire:
Discussing internally, we all agree you'd get bonus points if you had a follow up PR to add the struct name too as a property of the event in the case you fix the panic for here :)

PR #12, opened 2018-07-26
Record the struct name if the handler is a struct

apechimp:
As requested in #11 😉


nathanleclaire:
EXCELLENT! 20 points to (Whichever Harry Potter house you're in?)

I'll let @maplebed make the final call if we want to use `name` or another string here - it feels perhaps a bit precarious to have its meaning split across various uses, but I could be swayed either way. Thank you!

PR #13, opened 2018-08-07
add functions to marshal and unmarshal trace contxet to enable tracing

maplebed:
These headers match those in the node beeline for compatibility. See https://github.com/honeycombio/beeline-nodejs/blob/master/lib/propagation.js and the corresponding tests.

Test plan: go test. Compare tests to node, same tests passing should mean they're compatible.

### DIFF
--- a/internal/propagation.go
+++ b/internal/propagation.go
@@ -1,0 +1,116 @@
+package internal
+
+import (
+	"context"
+	"encoding/base64"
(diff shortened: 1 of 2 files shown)


maplebed:
closing - this was up there just to give an early feel for the code.  This stuff is now in the API branch where it can actually be used.

PR #14, opened 2018-08-08
Ben.api

maplebed:
Getting close. Would like reviews. I know it's a monstrous diff.
* beeline.go - public interface. Most code should exclusively write against this interface.
* internal/* - private methods intended for use by middleware authors
needs much additional testing and many additional tests.

### DIFF
--- a/beeline.go
+++ b/beeline.go
@@ -6,6 +6,7 @@ import (
 	"os"
 
 	internal "github.com/honeycombio/beeline-go/internal"
+	internal "github.com/honeycombio/beeline-go/internal/sample"
 	libhoney "github.com/honeycombio/libhoney-go"
 )
 
@@ -69,10 +70,9 @@ func Init(config Config) {
 		output = &libhoney.DiscardOutput{}
 	}
 	libhconfig := libhoney.Config{
-		WriteKey:   config.WriteKey,
-		Dataset:    config.Dataset,
-		SampleRate: config.SampleRate,
-		Output:     output,
+		WriteKey: config.WriteKey,
+		Dataset:  config.Dataset,
+		Output:   output,
 	}
 	if config.APIHost != "" {
 		libhconfig.APIHost = config.APIHost
@@ -95,6 +95,12 @@ func Init(config Config) {
 		// TODO add more debugging than just the responses queue
 		go readResponses(libhoney.Responses())
 	}
+
+	// configure and set a global sampler so sending traces can use it without
+	// threading it through
+	sampler := sample.NewDeterministicSampler(config.SampleRate)
+	sample.GlobalSampler = sampler
+
 	return
 }
 
@@ -168,6 +174,7 @@ func StartSpan(ctx context.Context) context.Context {
 	return internal.PushSpanOnStack(ctx)
 }
 
+// TODO change this to return a span object rather than a libhoney event to make it easier to treat appropriately for sampling
 // StartAsyncSpan is different from StartSpan in that it hands back a libhoney
 // event, and it's the caller's responsibility to handle everything from there
 // on - measuring the duration, adding appropriate fields, and sending the
--- a/beeline.go
+++ b/beeline.go
@@ -33,6 +33,16 @@ type Config struct {
 	// SamplRate is a positive integer indicating the rate at which to sample
 	// events. default: 1
 	SampleRate uint
+	// SamplerHook is a function that will get run with the contents of each
+	// event just before sending the event to Honeycomb. Register a function
+	// with this config option to have manual control over sampling within the
+	// beeline.
+	SamplerHook func(map[string]interface{}) (bool, int)
+	// PresendHook is a function call that will get run with the contents of
+	// each event just before sending them to Honeycomb. The function registered
+	// here may mutate the map passed in to add, change, or drop fields from the
+	// event before it gets sent to Honeycomb.
+	PresendHook func(map[string]interface{})
 	// APIHost is the hostname for the Honeycomb API server to which to send
 	// this event. default: https://api.honeycomb.io/
 	APIHost string
@@ -46,9 +56,6 @@ type Config struct {
 	// trouble getting the beeline to work, set this to true in a dev
 	// environment.
 	Debug bool
-
-	// disableTracing when set to true will suppress emitting trace.* fields
-	// disableTracing bool
 }
 
 // Init intializes the honeycomb instrumentation library.
@@ -101,6 +108,14 @@ func Init(config Config) {
 	sampler, err := sample.NewDeterministicSampler(config.SampleRate)
 	if err == nil {
 		sample.GlobalSampler = sampler
+	}
+
+	// pass through hooks if they're defined
+	if config.SamplerHook != nil {
+		internal.GlobalConfig.SamplerHook = config.SamplerHook
+	}
+	if config.PresendHook != nil {
+		internal.GlobalConfig.PresendHook = config.PresendHook
 	}
 	return
 }
--- a/beeline.go
+++ b/beeline.go
@@ -127,7 +127,7 @@ func Init(config Config) {
 // Flush implicitly ends all currently active spans.
 func Flush(ctx context.Context) {
 	trace := internal.GetTraceFromContext(ctx)
-	internal.SendTrace(trace)
+	trace.Send()
 	libhoney.Flush()
 }
 
@@ -168,9 +168,9 @@ func AddFieldToSpan(ctx context.Context, key string, val interface{}) {
 // fields on the root span are prefixed with `totals.app.`
 func AddRollupFieldToSpan(ctx context.Context, key string, val float64) {
 	namespacedKey := fmt.Sprintf("app.%s", key)
-	trace := internal.GetTraceFromContext(ctx)
-	if trace != nil {
-		trace.AddRollupField(namespacedKey, val)
+	span := internal.CurrentSpan(ctx)
+	if span != nil {
+		span.AddRollupField(namespacedKey, val)
 	}
 }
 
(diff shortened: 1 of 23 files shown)


beeline.go
```diff
@@ -34,6 +33,16 @@ type Config struct {
 	// SamplRate is a positive integer indicating the rate at which to sample
 	// events. default: 1
 	SampleRate uint
+	// SamplerHook is a function that will get run with the contents of each
+	// event just before sending the event to Honeycomb. Register a function
+	// with this config option to have manual control over sampling within the
+	// beeline. Runs before the PresendHook.
+	SamplerHook func(map[string]interface{}) (bool, int)
```
apechimp: 🤘

beeline.go
```diff
@@ -97,65 +102,136 @@ func Init(config Config) {
...
+}
+
+// TODO change this to return a span object rather than a libhoney event to make
+// it easier to treat appropriately for sampling
+
+// StartAsyncSpan is different from StartSpan in that it hands back a libhoney
+// event, and it's the caller's responsibility to handle everything from there
+// on - measuring the duration, adding appropriate fields, and sending the
+// event. Unlike normal spans, when finishing a trace, it does not get sent.
+func StartAsyncSpan(ctx context.Context, name string) *libhoney.Event {
```
apechimp: It'd be helpful to be able to create a background context with the event associated with an async span. This way we can just pass around a context instead of a context _and_ an event.

maplebed: Changed this to return a context, same as all the others. This should make it much more consistent to use.

internal/internal.go
```diff
@@ -2,201 +2,394 @@ package internal
...
+		amRoot:     true,
+		trace:      trace,
+		traceID:    traceID,
+		spanID:     spanID,
+		parentID:   parentID,
+		ev:         libhoney.NewEvent(),
+	}
+	span.ev.SampleRate = uint(trace.sampleRate)
+	span.ev.AddField("name", name)
+	span.ev.AddField("meta.root_span", true)
```
emfree: Say that service A starts a trace, and makes a call to service B, with trace headers. Service B calls `MakeNewTrace` with the IDs from the trace headers. Then the new span in service B has `meta.root_span: true`, even though it's not the global root span of the trace. It's only the root of the subtrace in service B. Do we care?

internal/propagation.go
```diff
@@ -0,0 +1,115 @@
...
+	}
+	return fmt.Sprintf(p.message, p.wrappedError)
+}
+
+func MarshalTraceContext(ctx context.Context) string {
+	currentSpan := CurrentSpan(ctx)
+
+	var prop = &Propagation{}
+	prop.Source = HeaderSourceBeeline
+	prop.TraceID = currentSpan.trace.headers.TraceID
```
emfree: Danger zone, can `currentSpan` be nil here?

internal/propagation.go
```diff
@@ -0,0 +1,115 @@
...
+}
+
+func (p *PropagationError) Error() string {
+	if p.wrappedError == nil {
+		return p.message
+	}
+	return fmt.Sprintf(p.message, p.wrappedError)
+}
+
+func MarshalTraceContext(ctx context.Context) string {
```
emfree: Might be nice to have `MarshalTraceContext`/`UnmarshalTraceContext` available to consumers of the public API. So that if someone needs to they can take the trace context and stuff it in a protobuf, or otherwise manage the propagation themselves.

beeline.go
```diff
@@ -97,65 +102,146 @@ func Init(config Config) {
...
 
 // Flush sends any pending events to Honeycomb. This is optional; events will be
-// flushed on a timer otherwies. It is useful to flush befare AWS Lambda
+// flushed on a timer otherwise. It is useful to flush before AWS Lambda
 // functions finish to ensure events get sent before AWS freezes the function.
-func Flush() {
+// Flush implicitly ends all currently active spans.
+func Flush(ctx context.Context) {
+	trace := internal.GetTraceFromContext(ctx)
+	trace.Send()
```
emfree: danger zone

beeline.go
```diff
@@ -97,65 +102,146 @@ func Init(config Config) {
...
+	return ctx
+}
+
+// FinishSpan finishes the currently active span. It should only be called for
+// spans created with StartTraceWithIDs, StartSpan, or StartAsyncSpan. Use the
+// context returned by finish span in order to ensure operations on the
+// "current" span continue to work correctly. The only time you should ignore
+// the context returned by this function is when using it as a defer to finish a
+// span also started in this scope; in that case when the function finishes the
+// previously scoped context will still have the right span marked as current.
```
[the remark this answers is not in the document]
maplebed: that's fine too. The problem comes when multiple spans are in the same function

```
func bigJob(ctx context.Context) {
  ctx = beeline.StartSpan("parentSpan", ctx)
  defer beeline.FinishSpan(ctx)
  ... do stuff ...
  ctx = beeline.StartSpan("childSpan", ctx)   // start child
  ... do stuff
  beeline.FinishSpan(ctx)                                // finish child
  ... do stuff
  beeline.AddFieldToSpan(ctx, "error", err) // Whoops! Adds "error" field to childSpan
}
```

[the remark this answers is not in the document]
emfree: In this case though, the better option is to give the child context a new name, instead of hazardously reusing `ctx`:

```
func bigJob(ctx context.Context) {
  ctx = beeline.StartSpan("parentSpan", ctx)
  defer beeline.FinishSpan(ctx)
  ... do stuff ...
  childCtx := beeline.StartSpan("childSpan", ctx)   // start child
  ... do stuff
  beeline.FinishSpan(childCtx)                     // finish child
  ... do stuff
  beeline.AddFieldToSpan(ctx, "error", err)
}
```

beeline.go
```diff
@@ -97,65 +102,146 @@ func Init(config Config) {
...
-	if bldr, ok := ctx.Value(honeyBuilderContextKey).(*libhoney.Builder); ok {
-		return bldr
+// AddFieldToTrace adds the field to both the currently active span and all
+// other spans involved in this trace that occur within this process.
+// Additionally, these fields are packaged up and passed along to downstream
+// processes if they are also using a beeline. This function is good for adding
+// context that is better scoped to the request than this specific unit of work,
+// eg user IDs, globally relevant feature flags, errors, etc. Fields added here
+// are prefixed with `global.`
+func AddFieldToTrace(ctx context.Context, key string, val interface{}) {
```
samstokes: I find the terminology here a bit awkward. As an application developer, the term "global" connotes _process-global_ (as in "global variable"), so I'd think of "global fields" as meaning "shared across _all requests_" (e.g. hostname, app configs). Whereas here we're defining it to mean "shared across all the spans in a _single request_".

I'd expect a good proportion of people to be confused by this (and, e.g., not realise they can add something like a team ID here, because it doesn't make sense for a team ID to be process-global. I'm also not sure the querying experience is very intuitive - `P95(durationMs) BREAK DOWN BY global.userID` would trip me up thinking "what does it mean to have a _global_ user ID?".

I'd prefer a prefix like `trace.` except of course that's reserved for trace meta-fields.

Ideas:
 * just use `app.`?
 * `request.`? Would only make sense for web apps, of course, and we might choose to worry about the potential for clobbering the fields from the request instrumentation.
 * what if whoever started the trace (by definition someone using the more-sophisticated API) picked a prefix for trace-level fields? e.g. an HTTP server middleware could set the prefix to be `request.`, a cron job wrapper or job queue consumer plugin could set it to be `job.`, etc.

beeline.go
```diff
@@ -97,65 +102,146 @@ func Init(config Config) {
...
-// contextBuilder isn't used yet but matches ContextEvent. When it's useful,
-// export it, but until then it's just confusing.
-func contextBuilder(ctx context.Context) *libhoney.Builder {
-	if bldr, ok := ctx.Value(honeyBuilderContextKey).(*libhoney.Builder); ok {
-		return bldr
+// AddFieldToTrace adds the field to both the currently active span and all
+// other spans involved in this trace that occur within this process.
+// Additionally, these fields are packaged up and passed along to downstream
+// processes if they are also using a beeline. This function is good for adding
+// context that is better scoped to the request than this specific unit of work,
```
samstokes: This is a bit confusing - haven't we previously used "processing a web request" as an example of what a unit of work _is_? More generally - what corresponds to a unit of work in the Beeline / tracing world? Is it a trace, a span, the root span specifically, or something else?

beeline.go
```diff
@@ -97,65 +102,146 @@ func Init(config Config) {
...
+// HasTrace returns true if there is a trace in the current context
+func HasTrace(ctx context.Context) bool {
+	trace := internal.GetTraceFromContext(ctx)
+	return trace != nil
+}
+
+// StartTraceWithIDs lets you start a trace with a specific set of IDs - it is
+// used when you've received the IDs from another source (eg incoming HTTP
+// headers). If you don't care what the IDs are, you may use either this or
+// StartSpan to start a trace. You cannot change the trace IDs after a trace has
```
samstokes: It's nice that `StartTraceWithIDs` calls `StartSpan` if needed and `StartSpan` calls `StartTraceWithIDs` if needed, but it leaves it a bit unclear which one you should call in the common case. Presumably the answer is that you should always call `StartSpan` unless you know what you're doing and want to start a new trace with a specific trace id? It'd be better to make that clearer (either by wording, or just by moving `StartTraceWithIDs` into the "advanced" API).

Also, "if you don't care what the IDs are" doesn't help me figure out _whether_ I care (or _should_ care). We could make this clearer by explicitly saying something like:

> Call `StartSpan` to start a new span. If there is no active trace, this will also start a new trace with an autogenerated unique trace id.

beeline.go
```diff
@@ -97,65 +102,146 @@ func Init(config Config) {
...
+// begun. StartTraceWithIDs also starts a span (the root span) for this trace.
+// You should call FinishSpan to close the span (and trace) started by this
+// function.
+func StartTraceWithIDs(ctx context.Context, traceID, parentID, name string) context.Context {
```
samstokes: What do you think about allowing this to take a map of fields as well? Given our trace context format includes fields to be propagated, it'd be nice if our function for starting a new trace supported those too, instead of having to do a separate `AddFieldToTrace` loop.

Unrelatedly, what happens if you call `StartTraceWithIDs` when there is already an active trace? I can't find the line reference now, but I think the node beeline reports an error in this case.

beeline.go
```diff
@@ -97,65 +102,146 @@ func Init(config Config) {
...
+// StartAsyncSpan is different from StartSpan in that when finishing a trace, it
+// does not get automatically sent. When finishing an async span it gets sent
+// immediately.
+func StartAsyncSpan(ctx context.Context, name string) context.Context {
+	ctx, _ = internal.StartAsyncSpan(ctx, name)
+	return ctx
+}
+
+// FinishSpan finishes the currently active span. It should only be called for
+// spans created with StartTraceWithIDs, StartSpan, or StartAsyncSpan. Use the
```
samstokes: It's a little unclear how to follow this advice ("`FinishSpan`... should only be called for spans created with ... `StartSpan`"). `FinishSpan` doesn't take a `span` parameter, so how am I supposed to know (besides diligence and code inspection) which span I'm finishing? The wording sort of implies that a) spans are things I can get my hands on (qv @emfree's comment), and b) there might be other ways to create spans besides `StartTraceWithIDs`, `StartSpan`, or `StartAsyncSpan`, but I should mistrust those other ways, or use some other way of finishing spans created thus.

Maybe clearer guidance would be "only call this after calling `StartSpan` etc, and don't call it more than once", or similar?

beeline.go
maplebed: the node beeline uses `app.` and doesn't differentiate (at the column name level) between span and trace level fields. I do appreciate that makes for one place to find things, but I also appreciate the signal that I can use this same field in filtering other types of spans.  For consistency, I'll go with `app.` to match the node beeline.

maplebed:
Closing in favor of #17

PR #17, opened 2018-08-21
Ben.api tracepkg

maplebed:
Biggest change - Trace and Span objects are pulled out of the `internal` package and into a `trace` package, intended to be used directly. Also created a `wrappers/common` package for things that all the HTTP and DB wrappers share. That felt better, because they're really specific to those wrappers rather than anything internal to the beeline itself.

This is ready to merge! I would like feedback on the overall shape as well as anything you pick out that I have missed, since I've been staring at this for too long.

### DIFF
--- a/README.md
+++ b/README.md
@@ -168,6 +168,7 @@ Available DB wrappers:
 
 * [`hnysql`](https://godoc.org/github.com/honeycombio/beeline-go/wrappers/hnysql) (for `database/sql`)
 * [`hnysqlx`](https://godoc.org/github.com/honeycombio/beeline-go/wrappers/hnysqlx) (for `github.com/jmoiron/sqlx`)
+* [`pop`](https://godoc.org/github.com/honeycombio/beeline-go/wrappers/pop) (for `github.com/gobuffalo/pop`)
 
 ## Other HTTP Frameworks
 
@@ -208,7 +209,7 @@ what's getting sent and modify your code to get what you would like to see.
 
 ### The events I'm sending aren't being accepted by Honeycomb
 
-Use the `PrintResponses` flag in configuring the beeline. This will print the
+Use the `Debug` flag in configuring the beeline. This will print the
 responses that come back from Honeycomb to the terminal when sending events.
 These responses will have extra detail saying why events are being rejected (or
 that they are being accepted) by Honeycomb.


trace/trace.go
```diff
@@ -0,0 +1,303 @@
...
+	AmAsync() bool
+	Finish()
+	GetParent() Span
+	ChildSpan(context.Context) (context.Context, Span)
+	ChildAsyncSpan(context.Context) (context.Context, Span)
+}
+
+// SyncSpan is the default span type.
+type SyncSpan struct {
+	amAsync      bool
```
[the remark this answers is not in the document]
maplebed: I have had severe side-eye at this type-level distinction ever since it showed up. Thanks for confirming; will expunge.

maplebed:
gnarly rebase. hope everything worked!

maplebed:
rebasey-basey

seanhagen:
Looks pretty awesome! Is the trace serialization something available in the other libraries as well? That would make propagating traces across PubSub so much easier.

beeline.go
```diff
@@ -178,3 +230,19 @@ func readResponses(responses chan libhoney.Response) {
 		}
 	}
 }
+
+// --------    remove from the main beeline package -----
```
samstokes: TODO?

doc.go
```diff
@@ -29,6 +36,9 @@
 //
 // There are runnable examples at
 // https://github.com/honeycombio/beeline-go/tree/master/examples and examples
-// of each wrapper in the dgodoc.
+// of each wrapper in the godoc.
 //
+// TODO create two comprehensive examples, one showing basic beeline use and the
```
samstokes: TODO?

trace/doc.go
```diff
@@ -0,0 +1,82 @@
...
+// used by the beeline.
+//
+// Summary
+//
+// Elementary use of the Honeycomb beeline has few needs - use a wrapper to
+// cover most of the basic info about a given event and then augment that event
+// with a few fields using `beeline.AddField`. Occasionally add an additional
+// span to the trace that is automatically created and add fields to that.
+// However, as the beeline starts being used in more sophisticated applications,
+// or new wrapers are being written, more direct control is needed to manipulate
```
samstokes: typo "wrapers" -> "wrappers"

trace/doc.go
```diff
@@ -0,0 +1,82 @@
...
+// for you, but if you're using the trace package directly you need to manage
+// that bookkeeping.
+//
+// When should you create a new span? There are no strict rules, but there are a
+// few heuristics. If there is a process that will repeat in a loop (a batch or
+// something) and each run through the loop is important, make a span. If the
+// code being instrumented has enough attributes that are relevant to it
+// directly to warrant its own bag of data, make a new span. If all you're
+// interested in is a simple timer, it's often cleaner to add that timer to the
+// current existing span instead and save the overhead of a new span.
```
samstokes: "Save the overhead of a new span" makes it sound like creating spans is expensive, in a "will this slow down my super fast Go app?" sort of way.

As I understand it, the distinction is basically whether you want to see the sub-operation in a trace. Am I right in thinking that creating a span will also give you timer-like semantics on the parent span via the rollup support?

trace/doc.go
```diff
@@ -0,0 +1,82 @@
...
+//
+// Sampling
+//
+// The default sampling applied by the beeline samples entire traces. For
+// example, if you set a sample rate to 10, then one out of 10 traces will be
+// sent, and all spans in that trace will be sent (or none at all). If you take
+// advantage of the SamplerHook, it is up to you and your implementation to
+// decide whether to sample entire traces or individual spans. If traces are
+// incomplete (i.e. some spans are kept and others dropped), the Honeycomb UI
+// will show missing traces where there are gaps.
```
samstokes: "missing traces where there are gaps" - should this read "missing spans" (or "missing parts of traces" or similar)?

trace/trace.go
```diff
@@ -0,0 +1,348 @@
...
+	"github.com/honeycombio/beeline-go/propagation"
+	"github.com/honeycombio/beeline-go/sample"
+	"github.com/honeycombio/beeline-go/timer"
+	libhoney "github.com/honeycombio/libhoney-go"
+)
+
+var GlobalConfig Config
+
+type Config struct {
+	// TODO describe what these are and the return values
```
samstokes: TODO?

trace/trace.go
```diff
@@ -0,0 +1,348 @@
...
+	PresendHook func(map[string]interface{})
+}
+
+// Trace holds some trace level state and the root of the span tree that will be
+// the entire in-process trace. Traces are sent to Honeycomb when the root span
+// is finished. You can send a trace manually, and that will cause all
+// synchronous  spans in the trace to be finished and sent. Asynchronous spans
+// must still be sent manually
+type Trace struct {
+	// TODO add shouldDrop
```
samstokes: TODO?

trace/trace.go
```diff
@@ -0,0 +1,348 @@
...
+// GetRootSpan returns the root of the in-process trace. Finishing the root span
+// will send the entire trace to Honeycomb. From the root span you can walk the
+// entire span tree using GetChildren.
+func (t *Trace) GetRootSpan() *Span {
+	return t.rootSpan
+}
+
+// Send will finish and send the all synchronous spans in the trace to Honeycomb
+func (t *Trace) Send() {
+	// TODO add sampling
```
samstokes: TODO?

trace/doc.go
```diff
@@ -0,0 +1,82 @@
...
+// are especially useful for goroutines that manage some background task after
+// the main user-facing chore has been completed. Examples of this are sending
+// an email or persisting accepted data.
+//
+// 2) creating traces with upstream parents when you are a downstream service.
+// The existing HTTP wrappers do this for you, but if your trace is getting
+// propagated via kafka or SQS or some other mechanism you may need to do this
+// yourself
+//
+// 3) adding fields that use a different naming scheme. Fields added via the
```
toshok: this feels like a bad idea when we depend on column names and their values for our server-side data model.

I'm not saying we'll see collisions immediately but we'll have to start actually mining customer dataset schemas to see if we will whenever we add a new beeline-specific column.

trace/doc.go
toshok: yeah, should be "missing spans" here.

that said, it's not always true - if leaf spans (or, e.g. an entire subtree) are dropped the UI can't/won't know.

trace/trace.go
```diff
@@ -0,0 +1,348 @@
...
+// get a field that represents the total time spent talking to the database from
+// all of the spans that are part of the trace.
+func (s *Span) AddRollupField(key string, val float64) {
+	s.rollupLock.Lock()
+	if s.rollupFields != nil {
+		s.rollupFields[key] += val
+	}
+	s.rollupLock.Unlock()
+	if s.trace != nil {
+		s.trace.rollupLock.Lock()
```
toshok: feels wrong to have spans locking the mutex in the trace.  Why not `Trace.addRollup(key, val)` and have the locks be local to the trace (as in `Trace.AddField`).

beeline.go
```diff
@@ -118,44 +150,64 @@ func Close() {
...
 // context from the request (`r.Context()`) and the key and value you wish to
-// add.
+// add.This function is good for span-level data, eg timers or the arguments to
+// a specific function call, etc. Fields added here are prefixed with `app.`
 func AddField(ctx context.Context, key string, val interface{}) {
-	ev := ContextEvent(ctx)
-	if ev == nil {
-		return
+	span := trace.GetSpanFromContext(ctx)
+	if span != nil {
```
toshok: what about the cases where Events are created?  should we not also add the field to the event?

trace/trace.go
```diff
@@ -0,0 +1,348 @@
...
+// AddRollupField adds a key/value pair to this span. If it is called repeatedly
+// on the same span, the values will be summed together.  Additionally, this
+// field will be summed across all spans and added to the trace as a total. It
+// is especially useful for doing things like adding the duration spent talking
+// to a specific external service - eg database time. The root span will then
+// get a field that represents the total time spent talking to the database from
+// all of the spans that are part of the trace.
+func (s *Span) AddRollupField(key string, val float64) {
+	s.rollupLock.Lock()
+	if s.rollupFields != nil {
```
toshok: why is it ever nil?  come to think of it, why is `s.rollupFields` a thing?  why not just `s.ev.AddField` for the local span, then make a call to the trace for rollup behavior?

trace/trace.go
```diff
@@ -0,0 +1,348 @@
...
+	}
+	s.ev.AddField("trace.span_id", s.spanID)
+	// add rollup fields to the event
+	for k, v := range s.rollupFields {
+		s.ev.AddField(k, v)
+	}
+	for _, child := range s.children {
+		if !child.AmAsync() {
+			if !child.amFinished {
+				child.AddField("meta.finished_by_parent", true)
```
toshok: I worry that this is not very discoverable.  That is, unless people notice this column/value, they might just think operations are taking longer than they should be.

version.go
```diff
@@ -1,3 +1,3 @@
 package beeline
 
-const version = "0.1.1"
+const version = "0.2.0"
```
toshok: :semver-sideeye: there are API breaks.  folks are using `vgo` with semver now, right?  probably want to bump this to 1.0.0.

trace/trace.go
```diff
@@ -0,0 +1,348 @@
...
+
+// Get Parent returns this span's parent.
+func (s *Span) GetParent() *Span {
+	return s.parent
+}
+
+// ChildAsyncSpan creates a child of the current span that is expected to
+// outlive the current span (and trace). Async spans must be manually sent using
+// the `Send()` method but are otherwise identical to normal spans.
+func (s *Span) ChildAsyncSpan(ctx context.Context) (context.Context, *Span) {
```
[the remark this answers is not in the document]
maplebed: Nathan objected to `New` in the name because it has strong meaning in normal go usage as a constructor (which this really isn't in the way that it would be in other Go things)

trace/trace.go
maplebed: rollup fields also sum if invoked multiple times on a span.

version.go
maplebed: it's pre-1.0, I'm allowed "beta" leeway. I'm not sure I'm quite comfortable calling this 1.0; if anything the previous one should be rebranded 1.0 and this 2.0.

maplebed:
@seanhagen The serialization format here is shared by the Node beeline: https://github.com/honeycombio/beeline-nodejs/blob/master/docs/API.md#interprocess-trace-propagation. Others will follow but aren't there yet.

trace/trace.go
```diff
@@ -0,0 +1,348 @@
...
+			spanType = "subroot"
+		}
+	case s.amAsync:
+		spanType = "async"
+	case len(s.children) == 0:
+		spanType = "leaf"
+	default:
+		spanType = "mid"
+	}
+	s.AddField("meta.span_type", spanType)
```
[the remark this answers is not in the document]
maplebed: I'm not sure. I didn't dig too hard, but I have a feeling the answer is no. The closest Open Tracing comes is to have a ChildOf relationship and a FollowsFrom relationship.

trace/trace.go
maplebed: 👍 added rollup method to the trace (and used it here) so its locks stay private.

doc.go
maplebed: yup, will leave to last though, because examples can come after landing.

version.go
toshok: yep, I am not quite up on my semver spec reading :)

https://semver.org/#spec-item-4

trace/trace.go
toshok: sigh, more un(der)documented behavioral differences :(

trace/doc.go
maplebed: removed the 'overhead of a new span part and just left it as unadorned advice.

trace/doc.go
maplebed: fixed it to say only middle spans that get dropped will show up as missing.

trace/trace.go
maplebed: done.

trace/trace.go
maplebed: I think premature optimization to think about `shouldDrop`. Removed the comment.

trace/trace.go
```diff
@@ -0,0 +1,348 @@
...
+	}
+	s.amFinished = true
+	// if we're closing the root span, send the whole trace
+	if s.amRoot {
+		s.trace.Send()
+	}
+}
+
+// AmAsync reveals whether the span is asynchronous (true) or synchronous (false).
+func (s *Span) AmAsync() bool {
```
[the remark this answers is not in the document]
maplebed: 👍

trace/trace.go
```diff
@@ -0,0 +1,348 @@
...
+// services so they may start a new trace that will be connected to this trace.
+// The serialized form may be passed to NewTrace() in order to create a new
+// trace that will be connected to this trace.
+func (s *Span) SerializeHeaders() string {
+	var prop = &propagation.Propagation{
+		TraceID:      s.trace.traceID,
+		ParentID:     s.spanID,
+		TraceContext: s.trace.traceLevelFields,
+	}
+	// prop.Source = HeaderSourceBeeline
```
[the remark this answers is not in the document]
maplebed: Eventually. Removed it for now. Idea is that if we can detect what style of headers we spy on the way in we should use them for outbound requests as well.

trace/trace.go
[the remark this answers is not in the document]
maplebed: Added to the column doc under 'lacks consensus' but will move if more beelines adopt.

beeline.go
maplebed: Events aren't created as part of the core beeline API. So nope?

trace/trace.go
maplebed: While true I think I need to punt fixing this to the server side. How can we better highlight errors and other unusual events that are showing up in a query?

version.go
maplebed: Though it is an interesting question - when should it be considered 1.0.0?

wrappers/common/common.go
```diff
@@ -0,0 +1,158 @@
...
+)
+
+type ResponseWriter struct {
+	http.ResponseWriter
+	Status int
+}
+
+func NewResponseWriter(w http.ResponseWriter) *ResponseWriter {
+	return &ResponseWriter{
+		ResponseWriter: httpsnoop.Wrap(w, httpsnoop.Hooks{}),
```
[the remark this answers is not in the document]
maplebed: It's supposed to help with #18 but ... I think I didn't do it right. Eventually that's what it's there for - to help implement all the interfaces instead of just the ResponseWriter.

wrappers/common/common.go
```diff
@@ -0,0 +1,158 @@
...
+	if args != nil {
+		ev.AddField("db.query_args", args)
+	}
+	return ev, fn
+}
+
+// BuildDBSpan does the same things as BuildDBEvent except that it has access to
+// a trace from the context and takes advantage of that to add the DB events
+// into the trace.
+func BuildDBSpan(ctx context.Context, bld *libhoney.Builder, query string, args ...interface{}) (context.Context, *trace.Span, func(error)) {
```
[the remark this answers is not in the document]
maplebed: 👍 pulled the shared bits into a third function

wrappers/hnysqlx/sqlx.go
```diff
@@ -75,7 +74,7 @@ func (db *DB) Beginx() (*Tx, error) {
 
 func (db *DB) BeginTxx(ctx context.Context, opts *sql.TxOptions) (*Tx, error) {
 	var err error
-	ev, sender := internal.BuildDBEvent(ctx, db.Builder, "")
+	ctx, span, sender := common.BuildDBSpan(ctx, db.Builder, "")
```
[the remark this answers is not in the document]
maplebed: hrmph. well I suppose this makes sense. It'll get a bit tricky to figure out sampling if you pass in an empty context to this stuff but I suppose that's something to figure out another time. I agree that it's unexpected that passing an empty context would result in no events at all rather than just trace-less events.

wrappers/common/common.go
```diff
@@ -0,0 +1,158 @@
...
+	timer := timer.Start()
+	parentSpan := trace.GetSpanFromContext(ctx)
+	if parentSpan == nil {
+		// if we have no trace or we're supposed to drop this trace, return a noop fn
```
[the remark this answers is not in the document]
maplebed: By folks not plumbing context all the way through from the initial request down to where the db is calling and then invoking a context-laden version with `context.Background()` or `context.TODO()` instead.

wrappers/hnynethttp/nethttp.go
```diff
@@ -69,29 +60,88 @@ func WrapHandler(handler http.Handler) http.Handler {
...
+	span := trace.GetSpanFromContext(ctx)
+	if span == nil {
+		// if there's no trace in the context, just send an event
+		tm := timer.Start()
+		ev := libhoney.NewEvent()
+		defer ev.Send()
+
+		// add in common request headers.
+		for k, v := range common.GetRequestProps(r) {
+			span.AddField(k, v)
```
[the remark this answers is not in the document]
maplebed: yup, typo, should be `ev.`

wrappers/hnynethttp/nethttp.go
```diff
@@ -69,29 +60,88 @@ func WrapHandler(handler http.Handler) http.Handler {
...
+	// we have a trace, let's use it and pass along trace context in addition to
+	// making a span around this HTTP call
+	ctx, span = span.ChildSpan(ctx)
+	defer span.Finish()
+	r = r.WithContext(ctx)
+	// add in common request headers.
+	for k, v := range common.GetRequestProps(r) {
+		span.AddField(k, v)
+	}
+	span.AddField("meta.type", "http_client")
```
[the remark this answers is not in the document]
maplebed: yeah, there is, but it's only a few lines on either side of the call to the wrapped RoundTripper. Too few to warrant another function in the same vein as with the db wrapping.  hrmph. Going to leave it but separate them into functions rather than a large if/else clause.

wrappers/hnynethttp/nethttp.go
```diff
@@ -69,29 +60,88 @@ func WrapHandler(handler http.Handler) http.Handler {
...
+type hnyTripper struct {
+	// wrt is the wrapped round tripper
+	wrt http.RoundTripper
+}
+
+func (ht *hnyTripper) RoundTrip(r *http.Request) (*http.Response, error) {
+	ctx := r.Context()
+	span := trace.GetSpanFromContext(ctx)
+	if span == nil {
+		// if there's no trace in the context, just send an event
```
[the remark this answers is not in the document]
maplebed: changed the place where it noops out.

trace/doc.go
maplebed: Part of the separation between the beeline / trace packages is that using the trace package gives you more flexibility (and therefore more possibility you break things). This part of the API is necessary for the wrappers (that create fields with `http.*` and `db.*`) to work without being extra privileged.  In order to make it possible for other people to write their own wrappers that are as effective as ours, this ability must be exposed somehow. This feels like the right place to put it.

trace/trace.go
```diff
@@ -0,0 +1,348 @@
...
+}
+
+// Send sends this span and any synchronous span children. Does not send any
+// async children. Primarily used on asynchronous spans - you must call `Send()`
+// on an asynchronous span in order for it to get sent to Honeycomb. Normally,
+// this function is not used for synchronous spans; they all get sent when the
+// root span gets `Finish()`ed. While you can call `Send` on a synchronous
+// (normal) span, doing so prevents the span from getting any trace level fields
+// added after it is sent.
+func (s *Span) Send() {
```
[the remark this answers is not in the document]
maplebed: I think I want to leave the behavior as is for the time being and have updated the function description to be a little more explicit. I think it will be worth revisiting if we decide to go down the delayed send route for full traces.

trace/trace.go
maplebed: done.

maplebed:
I think I have addressed all of Sam's comments and most of Toshok's comments. There's always more documentation and example work to do, but i think it's back ready for another round.

trace/doc.go
toshok: Sure, there needs to be two paths, but we can have an api that users use and an api that instrumentations use.  I don't think we should have a single api that both use that permits collisions like this.

beeline.go
toshok: going to chalk this up to confusion about wrappers creating both events and spans.

I haven't gone back and audited, but do all wrappers that send events send them from the method where they're created?  is that always going to be the case?

PR #19, opened 2018-08-27
Add X-Forwarded-For and X-Forwarded-Proto headers.

dynamike:
Many proxies support the `X-Forwarded-For` and `X-Forwarded-Proto` HTTP headers. This PR pulls them in as `request.header` fields. I'm slightly ambivalent if this is generally applicable for beeline-go, but I will leave it to you folks to figure that out. 

**Testing**
- [x] `go test` passes
- [x]  Verify examples correctly add `request.header.x_forwarded_for` and `request.header.x_forwarded_proto` fields.


internal/internal.go
```diff
@@ -43,6 +43,8 @@ func AddRequestProps(req *http.Request, ev *libhoney.Event) {
 	ev.AddField("request.content_length", req.ContentLength)
 	ev.AddField("request.remote_addr", req.RemoteAddr)
 	ev.AddField("request.header.user_agent", req.UserAgent())
+	ev.AddField("request.header.x_forwarded_for", req.Header.Get("x-forwarded-for"))
```
maplebed: I'd like to see these each surrounded by a conditional to only be added if they're not the empty string. While `req.Header.Get()` is kind to missing keys (when the headers are absent), it's better from the Honeycomb dataset perspective to not report the field rather than to report an empty field. 
Come to think of it, that probably should apply to user_agent as well, though I expect that being empty is more a rare exception rather than as common as folks not being in AWS (and so not having the forwarded_proto header, for example).

dynamike:
@maplebed Good call on the conditionals. The PR is updated. Let me know if you want anything else.

PR #20, opened 2018-08-30
Use passed in context to build event

seanhagen:
Using `context.Background()` in these functions was causing these trace spans to
become disconnected from their parent spans.


maplebed:
Thank you! You're right this is clearly a mistake.

PR #21, opened 2018-08-30
Commit with context

seanhagen:
Add a method so that when a transaction is comitted, the commit shows up as part of the trace instead of being orphaned.

### DIFF
--- a/wrappers/hnysqlx/sqlx.go
+++ b/wrappers/hnysqlx/sqlx.go
@@ -1842,6 +1842,21 @@ func (tx *Tx) Rollback() error {
 	return err
 }
 
+func (tx *Tx) Rollbackx(ctx context.Context) error {
+	var err error
+	_, sender := internal.BuildDBEvent(ctx, tx.Builder, "")
+	defer sender(err)
+
+	// ensure any changes to the Mapper get passed along
+	if tx.Mapper != nil {
+		tx.wtx.Mapper = tx.Mapper
+	}
+
+	// do DB call
+	err = tx.wtx.Rollback()
+	return err
+}
+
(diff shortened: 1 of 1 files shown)


maplebed:
Ooh, I hadn't considered adding additional methods beyond what sql and sqlx expose, but for someone who's all in on the beeline, I like it! Would it also be useful to add `Rollbackx()`?

seanhagen:
Yeah, `Rollbackx` is a good idea -- don't want those to be orphaned either! I'll add it.

nathanleclaire:
Cool! My 2c of naming bikeshed would be that I think I'd prefer to see `CommitContext` and `RollbackContext` instead of `Commitx` ... following the example set by `Dial` and `DialContext` in the Go standard library (https://tip.golang.org/pkg/net/#Dialer.Dial). My reasoning for this is that `Queryx` is very specifically a `sqlx` thing, and while I'm in favor of introducing a Beeline-specific function not present in the wrapped library, it should ideally not imply so strongly that it comes from the underlying package.

maplebed:
@nathanleclaire has a good point - the `x` in other `sqlx` methods refers to the fact that it returns a `sqlx` object rather than a `sql` object. Appending `Context` to the name is saying "this is a context-laden version of the command". 

(example: [Query](https://godoc.org/github.com/jmoiron/sqlx#NamedStmt.Query) vs. [Queryx](https://godoc.org/github.com/jmoiron/sqlx#NamedStmt.Queryx))

seanhagen:
Makes sense, I'll change it.

nathanleclaire:
👍 😎 👍 LGTM

Thank you for the contributions!

PR #23, opened 2018-09-16
Prevent obscuring optional http.ResponseWriter methods.

### LINKED ISSUE #18: internal.ResponseWriter doesn't implement http.Hijacker
Hello! Thanks for the Go libraries – they're very useful.

In our API server, we create a `gorilla.Mux` that routes to both HTTP endpoints and Websocket endpoints. The Websocket endpoints use a [`websocket.Upgrader`](https://godoc.org/github.com/gorilla/websocket#Upgrader.Upgrade).

Unfortunately, Beeline's `internal.ResponseWriter` doesn't implement `http.Hijacker`, which the websocket upgrader requires. So we can't use Beeline for those endpoints.

We're working around the issue by only piping the HTTP handlers through the supplied Gorilla middleware, but it would be great if it worked (or at least didn't break) our websocket endpoints.

We've previously used https://github.com/felixge/httpsnoop to do this for our websocket endpoints, and it looks like you used to also. Though I understand why you might want to reduce your number of third-party dependencies.

Thanks again!

vcabbage:
* Use the ResponseWriter returned from httpsnoop.Wrap so that optional methods are still exposed on the concrete type.
* Only record the first set status code.

I'm not sure `ResponseWriter` is a good name for the struct anymore. I considered renaming it to `ResponseRecorder` but didn't want to make too many changes.

Resolves #18

### DIFF
--- a/wrappers/common/common.go
+++ b/wrappers/common/common.go
@@ -14,19 +14,31 @@ import (
 )
 
 type ResponseWriter struct {
-	http.ResponseWriter
-	Status int
+	// Wrapped is not embedded to prevent ResponseWriter from directly
+	// fulfilling the http.ResponseWriter interface. Wrapping in this
+	// way would obscure optional http.ResponseWriter interfaces.
+	Wrapped http.ResponseWriter
+	Status  int
 }
 
 func NewResponseWriter(w http.ResponseWriter) *ResponseWriter {
-	return &ResponseWriter{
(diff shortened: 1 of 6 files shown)


maplebed:
Thanks for the PR! I'm sorry I haven't gotten to review it yet - it's high on the list for next week. Just wanted to provide an update rather than silence.

toshok:
@vcabbage taking a look :)

toshok:
🎉

vcabbage:
Thanks!

PR #24, opened 2018-09-19
fix rollup - fields should be added to the root span

maplebed:
Added test confirming that rolled up fields actually get set on the root span.  They didn't. Fixed the code so they do. Tests pass.  Yay!

### DIFF
--- a/trace/trace.go
+++ b/trace/trace.go
@@ -213,7 +213,7 @@ func (s *Span) Send() {
 		s.ev.AddField("trace.parent_id", s.parentID)
 	}
 	s.ev.AddField("trace.span_id", s.spanID)
-	// add rollup fields to the event
+	// add this span's rollup fields to the event
 	for k, v := range s.rollupFields {
 		s.ev.AddField(k, v)
 	}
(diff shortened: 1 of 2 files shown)


tredman:
Do I understand this right? If I'm going through and calling `AddRollupField` like this:
```
root span:
  AddRollupSpan("foo", 1)
s1:
  AddRollupSpan("foo", 1)
s2:
  AddRollupSpan("foo", 1)
```

I should get three events back with fields:
```
root: { "foo": 1, "rollup.foo": 3}
s1: {"foo": 1 }
s2: {"foo": 1}
```
It seems ambiguous to me to have a rollup field on my span called `foo` that I know is a rollup field (because I remember calling `AddRollupField`!), and then another that is explicitly prefixed with `rollup` - can we be more clear that this is a rollup across the entire trace? Something like `global_rollup` or `tracewide_rollup`?

maplebed:
I think the were originally called global (err... maybe it was trace level fields that were called global...), but that got squashed because in code it makes it feel like it's global to the process not the trace.  

The expected use case is when you have a thing that happens a bunch in the lower spans that you want a summary in the root span. It would be unusual to call AddRollupField in the root span. The way it's used in the beeline is in the sql wrapper - each database call has a field `db.duration_ms` that says how long _that_ database call took, and then the root span gets a field called `rollup.db.duration_ms` that says how long _all_ the database calls spawned from this root took. 

If possible I think I'd prefer to keep the field name the more terse version and push more clarity into the documentation. (which is certainly reasonable; it's one of the least documented calls in the beeline). Current documentation: https://godoc.org/github.com/honeycombio/beeline-go/trace#Span.AddRollupField

tredman:
Ok fair enough. Yeah some more notes in the docs about this would help.

Issue #32, opened 2018-10-25
Race in sampler hook


apechimp:
When we switched to the new tracing stuff introduced in `v0.2.x`, we started seeing panics like the following.
```Oct 25 10:27:49 ip-10-10-1-212 b94ce8 syslog fatal fatal error: concurrent map iteration and map write
Oct 25 10:27:49 ip-10-10-1-212 b94ce8 syslog  goroutine 64181 [running]:
Oct 25 10:27:49 ip-10-10-1-212 b94ce8 syslog  runtime.throw(0x12ac57f, 0x26)
Oct 25 10:27:49 ip-10-10-1-212 b94ce8 syslog  #011/usr/local/go/src/runtime/panic.go:605 +0x95 fp=0xc51d7eee90 sp=0xc51d7eee70 pc=0x42dc45
Oct 25 10:27:49 ip-10-10-1-212 b94ce8 syslog  runtime.mapiternext(0xc51d7ef030)
Oct 25 10:27:49 ip-10-10-1-212 b94ce8 syslog  #011/usr/local/go/src/runtime/hashmap.go:778 +0x6f1 fp=0xc51d7eef28 sp=0xc51d7eee90 pc=0x40bad1
Oct 25 10:27:49 ip-10-10-1-212 b94ce8 syslog  github.com/launchdarkly/event-recorder/vendor/github.com/launchdarkly/foundation/honey.SamplerHook(0xc54581ffb0, 0x12901de, 0xe)
Oct 25 10:27:49 ip-10-10-1-212 b94ce8 syslog  #011/go/src/github.com/launchdarkly/event-recorder/vendor/github.com/launchdarkly/foundation/honey/honeycomb.go:35 +0x127 fp=0xc51d7ef130 sp=0xc51d7eef28 pc=0xecfc17
Oct 25 10:27:49 ip-10-10-1-212 b94ce8 syslog  github.com/launchdarkly/event-recorder/vendor/github.com/honeycombio/beeline-go/trace.(*Span).send(0xc4896de500)
Oct 25 10:27:49 ip-10-10-1-212 b94ce8 syslog  #011/go/src/github.com/launchdarkly/event-recorder/vendor/github.com/honeycombio/beeline-go/trace/trace.go:328 +0x3c6 fp=0xc51d7ef278 sp=0xc51d7ef130 pc=0x753f66
[...]

PR #33, opened 2018-10-25
[trace] Add lock in AddField and Send to prevent panic due to race in…

tredman:
… map access

Fixes https://github.com/honeycombio/beeline-go/issues/32

I was able to repro this race condition reliably with `TestAddFieldDoesNotCauseRaceInSendHooks`. Adding a new mutex takes care of the issue, although we might want a better way to handle locking going forward. Libhoney's `Fields()` function should not return a reference to its internal map without also exposing some kind of locking mechanism to the caller.

### DIFF
--- a/trace/trace.go
+++ b/trace/trace.go
@@ -142,6 +142,7 @@ type Span struct {
 	rollupLock   sync.Mutex
 	timer        timer.Timer
 	trace        *Trace
+	eventLock    sync.Mutex
 }
 
 // newSpan takes care of *some* of the initialization necessary to create a new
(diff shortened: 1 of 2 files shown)


trace/trace.go
```diff
@@ -321,6 +326,12 @@ func (s *Span) send() {
 		}
 	}
 
+	// Because we hand a raw map over to the Sampler and Presend hooks, it's
+	// possible for the user to modify/iterate over the map in these hooks and
+	// still modify the event somewhere else with AddEvent. We lock here to
```
emfree: nit: s/AddEvent/AddField ?

trace/trace.go
```diff
@@ -321,6 +326,12 @@ func (s *Span) send() {
 		}
```
emfree: it looks like there are a few places here and in the `Send()` method where we call `s.ev.AddField` instead of `s.AddField`. This is fine currently, but it might be worth converting to `s.AddField` throughout to avoid potentially introducing similar race conditions in the future?

Issue #38, opened 2018-11-21
Data race when sending spans 


blake-wilson:
While running some tests with the race detector enabled, we noticed the following data race in the `Span`'s `Send` method:

```
Read at 0x00c008aed051 by goroutine 173:
  github.com/{***}/vendor/github.com/honeycombio/beeline-go/trace.(*Span).Send()
      /go/src/github.com/{***}/vendor/github.com/honeycombio/beeline-go/trace/trace.go:241 +0x3f6
  github.com/{***}/vendor/github.com/{***}/finstrument.(*Span).Finish()

Previous write at 0x00c008aed051 by goroutine 212:
  github.com/{***}/vendor/github.com/honeycombio/beeline-go/trace.(*Span).Send()
      /go/src/github.com/{***}/vendor/github.com/honeycombio/beeline-go/trace/trace.go:250 +0x48c
  github.com/{***}/vendor/github.com/{***}/finstrument.(*Span).Finish()
```

I am using the latest master of libhoney-go (a6839afb6be6f3ec9de43b3a2c08f89b338e8534)  and beeline-go (87bb646f47ddd2e7226aba28e83bd08c2c821e17).

These lines correspond to the following code 
- https://github.com/honeycombio/beeline-go/blob/master/trace/trace.go#L241
- https://github.com/honeycombio/beeline-go/blob/master/trace/trace.go#L250

It's not clear exactly what happened to me, but perhaps if a parent span is sent at the same time as one of its children it could lead to this race?

PR #40, opened 2018-11-23
[trace] fix race when child and parent send is called at the same time

tredman:
Fixes https://github.com/honeycombio/beeline-go/issues/38  - if `Send` is called for a span and its parent at the same time, a race can occur when accesing "isSent". This adds a lock to `Send`, which sets `isSent` and protects reads with read locks. This also effectively prevents `Send` from executing multiple times for the same span.

### DIFF
--- a/trace/trace.go
+++ b/trace/trace.go
@@ -270,24 +270,13 @@ func (s *Span) GetParent() *Span {
 // outlive the current span (and trace). Async spans are not automatically sent
 // when their parent finishes, but are otherwise identical to synchronous spans.
 func (s *Span) CreateAsyncChild(ctx context.Context) (context.Context, *Span) {
-	ctx, newSpan := s.CreateChild(ctx)
-	newSpan.isAsync = true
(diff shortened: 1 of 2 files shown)


trace/trace.go
```diff
@@ -238,16 +246,20 @@ func (s *Span) Send() {
 	s.childrenLock.Lock()
 	for _, child := range s.children {
 		if !child.IsAsync() {
-			if !child.isSent {
+			child.sendLock.RLock()
+			isChildSent := child.isSent
+			child.sendLock.RUnlock()
+			if !isChildSent {
 				child.AddField("meta.sent_by_parent", true)
```
martin308: it seems like it could still be possible for this field to be set on a child which was not sent by a parent? I'm not sure how important that is if it possible though 🤔

tredman: Yeah this is not perfect, but holding the lock here will deadlock the `Send` method. I think it might require a new method that sets the field while under the send lock.

Issue #42, opened 2019-01-04
Memory usage is high for long (duration) traces


apechimp:
We are using tracing to understand the lifecycle of connections that can be very long lived. The way the trace package is designed, spans continue to be referenced by the root span until the root span is sent. As a result, spans end up consuming memory even after they have been sent.

maplebed:
I think it may be possible to nil out the references that link spans together into a tree as each span gets sent, which would allow the garbage collector to reclaim that memory.

apechimp:
I put up an attempt at what you suggested. I'm also realizing that you run into a similar problem with async spans and their parent.

PR #43, opened 2019-01-05
Delete spans from children array as they're sent

apechimp:
This way they can be GC'd if they're from a root span which is around for a long time.

Resolves https://github.com/honeycombio/beeline-go/issues/42.

### DIFF
--- a/trace/trace.go
+++ b/trace/trace.go
@@ -243,22 +243,44 @@ func (s *Span) Send() {
 		s.AddField(k, v)
 	}
 	s.rollupLock.Unlock()
+
 	s.childrenLock.Lock()
+	childrenToSend := make([]*Span, 0, len(s.children))
 	for _, child := range s.children {
 		if !child.IsAsync() {
 			child.sendLock.RLock()
 			isChildSent := child.isSent
 			child.sendLock.RUnlock()
 			if !isChildSent {
-				child.AddField("meta.sent_by_parent", true)
-				child.Send()
+				// queue children up to be sent. We'd deadlock if we actually sent the
+				// child here.
+				childrenToSend = append(childrenToSend, child)
 			}
 		}
 	}
 	s.childrenLock.Unlock()
 
+	for _, child := range childrenToSend {
+		child.AddField("meta.sent_by_parent", true)
+		child.Send()
+	}
+
 	s.send()
 	s.isSent = true
+
+	// Remove this span from its parent's children list so that it can be GC'd
+	if s.parent != nil {
+		s.parent.childrenLock.Lock()
+		children := make([]*Span, 0, len(s.parent.children)-1)
+		for _, child := range s.parent.children {
+			if child == s {
+				continue
+			}
+			children = append(children, child)
+		}
+		s.parent.children = children
+		s.parent.childrenLock.Unlock()
+	}
 
 }
 
--- a/trace/trace_test.go
+++ b/trace/trace_test.go
@@ -135,6 +135,12 @@ func TestSpan(t *testing.T) {
 	assert.NotNil(t, span.timer, "span should have an initialized timer")
 	assert.Equal(t, tr, span.trace, "span should have a pointer to trace")
 
+	_, childSpan := span.CreateChild(ctx)
+	assert.Equal(t, span.spanID, childSpan.parentID, "child span's parent ID should be parent's span ID")
+	assert.Equal(t, span, childSpan.parent, "child span should have a pointer to parent")
+	assert.Len(t, span.children, 1, "parent span should have a child")
+	assert.Equal(t, childSpan, span.children[0], "parent span's child should be child span")
+
 	ctx, asyncSpan := rs.CreateAsyncChild(ctx)
 	assert.Equal(t, true, asyncSpan.isAsync, "async span should not be async")
 	assert.Equal(t, true, asyncSpan.IsAsync(), "async span should not be async")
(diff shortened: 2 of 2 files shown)


maplebed:
I like this, and I need to think through a few things. I'm struggling to wrap my head around it enough to answer any of these questions.

* Before this change, starting from the trace and the root span, you can walk the entire tree. With this change, you can't - any span that gets spent will break the tree. Does that matter? Does anything rely on walking the tree?

* Since every span gets sent, the `children` list will have to grow and shrink one span at a time. Is this a big deal? Would it make a difference if once children are sent their entry in the list got replaced by `nil` instead of shortening the list?

* Is there anything about async children that should be treated differently from a normal span? In other words, are there any considerations around when a span gets sent in relationship to when its children get sent that would come in to play here?

* This change only removes a span from it's parent's child list when it gets sent.  Should it also zero out it's own parent reference so as to break the tree in both directions?

* Specifically from a code perspective, `s.parent.removeSentSpan(s)` feels a little awkward, though I understand why it makes sense to keep locks within the ownership of the object that has them. Is there some way to wrap everything in to a `s.remove()` to make it clear that it's this span that's going away? i.e. should a span remove itself or should the removal be under the parent's control? (or both?) Maybe I'd be happier if the function name was more explicit about being for deleting a child (`removeChildSpan(s)`? `removeSentChild(s)`?)

* Is this removal something that should be entirely private to the beeline or is there any reason it might be useful to export the method and expose it as part of the public API? Is there ever a reason to delete an unsent span?

I think this change is the right thing to do, I'm just a bit nervous about it complicating the internal data structure in some way that is not obvious and yet somehow subtly harmful.

apechimp:
We discussed this AFK a bit, but I'll reiterate here for posterity.

1. I am not sure about anything using the tree. I've been running this branch for a few days with a pretty good workload and it seems to be OK. Since spans can only be sent once, it seems like it could be problematic to change them post ex facto.
2. I'll write some benchmarks here. We don't want to `nil` out because we could allocate an arbitrarily large array with long traces. An alternative would be to check `isSent`, but that would require a lock.
3. I don't think this is affected, but I do wonder about including async spans in the parent's children list at all. It seems like you wouldn't ever need to inspect them from the parent...
4. That's correct. I think the parent case is a little easier though. If the span is __not__ async, the only reason you'd want to remove the parent reference is if the parent span has been sent. However, if the parent span has been sent, the child span has been sent as well, so you're OK: the GC can't reach it. If the span _is_ async, I think we might want to avoid setting the `parent` altogether. AFAICT, the only time the parent is used is in the code I've added here. There's also `GetParent`, but I've never wanted to use that, so I can't really comment on that.
5. I think renaming to `removeChildSpan(s)` is probably best. That's much clearer.
6. I think it should be private (at least for now) I can't really think of any reason you'd want to remove a span from a trace other than when it has been sent.

I can rename that method and benchmark to see how different removal strategies work. I think the parent stuff is needed too, but should probably be a separate PR.

apechimp:
I renamed that method and added a benchmark. The benchmark results follow.

master
```
$ go test -benchmem -count 10 -bench SendChildSpans -run nothing ./trace
goos: darwin
goarch: amd64
pkg: github.com/honeycombio/beeline-go/trace
BenchmarkSendChildSpans-8        1000000              1668 ns/op            1027 B/op         15 allocs/op
BenchmarkSendChildSpans-8        1000000              1463 ns/op            1027 B/op         15 allocs/op
BenchmarkSendChildSpans-8        1000000              1549 ns/op            1027 B/op         15 allocs/op
BenchmarkSendChildSpans-8        1000000              1476 ns/op            1027 B/op         15 allocs/op
BenchmarkSendChildSpans-8        1000000              1525 ns/op            1027 B/op         15 allocs/op
BenchmarkSendChildSpans-8        1000000              1478 ns/op            1027 B/op         15 allocs/op
BenchmarkSendChildSpans-8        1000000              1579 ns/op            1027 B/op         15 allocs/op
BenchmarkSendChildSpans-8        1000000              1529 ns/op            1027 B/op         15 allocs/op
BenchmarkSendChildSpans-8        1000000              1522 ns/op            1027 B/op         15 allocs/op
BenchmarkSendChildSpans-8        1000000              1542 ns/op            1027 B/op         15 allocs/op
PASS
ok      github.com/honeycombio/beeline-go/trace 17.574s
```

clean-up-sent-spans
```
$ go test -benchmem -count 10 -bench SendChildSpans -run nothing ./trace
goos: darwin
goarch: amd64
pkg: github.com/honeycombio/beeline-go/trace
BenchmarkSendChildSpans-8        1000000              1684 ns/op            1009 B/op         18 allocs/op
BenchmarkSendChildSpans-8        1000000              1562 ns/op            1008 B/op         18 allocs/op
BenchmarkSendChildSpans-8        1000000              1591 ns/op            1007 B/op         18 allocs/op
BenchmarkSendChildSpans-8        1000000              1564 ns/op            1005 B/op         17 allocs/op
BenchmarkSendChildSpans-8        1000000              1528 ns/op            1006 B/op         18 allocs/op
BenchmarkSendChildSpans-8        1000000              1556 ns/op            1006 B/op         18 allocs/op
BenchmarkSendChildSpans-8        1000000              1580 ns/op            1007 B/op         18 allocs/op
BenchmarkSendChildSpans-8        1000000              1564 ns/op            1007 B/op         18 allocs/op
BenchmarkSendChildSpans-8        1000000              1549 ns/op            1006 B/op         18 allocs/op
BenchmarkSendChildSpans-8        1000000              1559 ns/op            1005 B/op         17 allocs/op
PASS
ok      github.com/honeycombio/beeline-go/trace 17.826s
```

We're looking at a couple of extra allocations and _slightly_ more time for sending.

PR #45, opened 2019-01-08
add url queries, add name even when empty

maplebed:
There was nothing in the beeline that would show you the full URL requested. This adds that under `request.url` and also adds just the query portion of the URL at `request.query`. The URL is close to path + query, but is actually more - anchors (#foo) will appear in the URL but aren't explicitly called out into their own field here.

### DIFF
--- a/wrappers/hnynethttp/nethttp.go
+++ b/wrappers/hnynethttp/nethttp.go
@@ -45,6 +45,9 @@ func WrapHandler(handler http.Handler) http.Handler {
 			if handlerName != "" {
 				span.AddField("handler.name", handlerName)
 				span.AddField("name", handlerName)
+			} else {
+				// we always want a name, even if it's kinda useless.
+				span.AddField("name", "handler")
 			}
 		}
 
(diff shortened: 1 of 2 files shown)


wrappers/hnynethttp/nethttp.go
```diff
@@ -45,6 +45,9 @@ func WrapHandler(handler http.Handler) http.Handler {
 			if handlerName != "" {
 				span.AddField("handler.name", handlerName)
 				span.AddField("name", handlerName)
+			} else {
+				// we always want a name, even if it's kinda useless.
+				span.AddField("name", "handler")
```
toshok: thoughts on just always sending `name: "handler"` (or `"http handler"` or something) and then only sending `handler.name: handlerName` if handlerName has a value?

maplebed: I think there's a lot of value in having a bit more detail in the `name` field; having it always be "handler" feels more like a `type` than a `name`.

PR #48, opened 2019-02-06
add a convenience function to copy the trace bits from one context to another

maplebed:
When breaking context to isolate a section of code from a timeout or cancellation in its parent, it's necessary to create a new context for the isolated code. Copying the context bits that are necessary to continue the beeline trace is annoying; this function makes it a little bit easier.

### DIFF
--- a/trace/context.go
+++ b/trace/context.go
@@ -3,8 +3,6 @@ package trace
 import (
 	"context"
 	"errors"
-
-	"github.com/honeycombio/beeline-go/trace"
 )
 
 const (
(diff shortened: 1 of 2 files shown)


trace/context.go
```diff
@@ -48,3 +55,20 @@ func GetSpanFromContext(ctx context.Context) *Span {
 func PutSpanInContext(ctx context.Context, span *Span) context.Context {
 	return context.WithValue(ctx, honeySpanContextKey, span)
 }
+
+// CopyContext takes a context that has a beeline trace and one that doesn't. It
+// copies all the bits necessary to continue the trace from one to the other.
+// This is useful if you need to break context to launch a goroutine that
+// shouldn't be cancelled by the parent's cancellation context. It returns the
+// newly populated context. If it can't find a trace in the source context, it
+// returns the unchanged dest context with an error source context.
```
ianwilkes: "with an error source context." ?

I think you just mean "with an error."

PR #50, opened 2019-02-20
Enable setting BlockOnSend when configuring the beeline

tredman:
We enable this setting in libhoney but don't allow a way to configure it via the beeline.

### DIFF
--- a/beeline.go
+++ b/beeline.go
@@ -68,8 +68,13 @@ type Config struct {
 	// (50) that are sent per batch.
 	MaxBatchSize uint
 	// PendingWorkCapacity overrides the default event queue size (1000).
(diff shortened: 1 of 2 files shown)


maplebed:
libhoney 1.9.0 is nearly released, which will let you configure a libhoney Client with all the necessary settings. Rather than duplicating some of the list here, I'd like to add an option to configure the beeline with a custom libhoney Client and keep all the settings there. This is both more concise (one setting instead of many) and more complete (even adding `BlockOnSend` here, there are still more configuration options that still can't be set). 

I suggest closing this one without merging and _soon_ there'll be a similar one that allows the client to be set.

PR #52, opened 2019-03-12
New middleware for Echo router

jamietsao:
Added a honeycomb instrumentation for the Echo router (https://echo.labstack.com/) via middleware

### DIFF
--- a/wrappers/hnyecho/echo.go
+++ b/wrappers/hnyecho/echo.go
@@ -11,8 +11,7 @@ import (
 type (
 	EchoWrapper struct {
 		handlerNames map[string]string
-
-		once sync.Once
+		once         sync.Once
 	}
 )
 
@@ -40,9 +39,12 @@ func (e *EchoWrapper) Middleware() echo.MiddlewareFunc {
 			span.AddField("handler.name", handlerName)
 			span.AddField("name", handlerName)
 
-			// add field for each path param
+			// add route related fields
+			span.AddField("route", c.Path())
+			span.AddField("route.handler", handlerName)
 			for _, name := range c.ParamNames() {
-				span.AddField("app."+name, c.Param(name))
+				// add field for each path param
+				span.AddField("route.params."+name, c.Param(name))
 			}
 
 			// invoke next middleware in chain
--- a/wrappers/hnyecho/echo.go
+++ b/wrappers/hnyecho/echo.go
@@ -44,8 +44,9 @@ func (e *EchoWrapper) Middleware() echo.MiddlewareFunc {
 			// invoke next middleware in chain
 			err := next(c)
 
-			// add field for http response code
+			// add fields for http response code and size
 			span.AddField("response.status_code", c.Response().Status)
+			span.AddField("response.size", c.Response().Size)
 
 			return err
 		}
--- a/wrappers/hnyecho/echo.go
+++ b/wrappers/hnyecho/echo.go
@@ -1,7 +1,7 @@
 package hnyecho
 
 import (
-	"github.com/jamietsao/beeline-go/wrappers/common"
+	"github.com/honeycombio/beeline-go/wrappers/common"
 	"github.com/labstack/echo"
 )
 
--- a/wrappers/hnyecho/echo.go
+++ b/wrappers/hnyecho/echo.go
@@ -1,6 +1,8 @@
 package hnyecho
 
 import (
+	"sync"
+
 	"github.com/honeycombio/beeline-go/wrappers/common"
 	"github.com/labstack/echo"
 )
@@ -9,6 +11,8 @@ import (
 type (
 	EchoWrapper struct {
 		handlerNames map[string]string
+
+		once sync.Once
 	}
 )
 
@@ -55,18 +59,19 @@ func (e *EchoWrapper) Middleware() echo.MiddlewareFunc {
 
 // Unfortunately the name of c.Handler() is an anonymous function
 // (https://github.com/labstack/echo/blob/master/echo.go#L487-L494).
-// This function will build a map of request paths to actual handler names during
-// the first request, thus providing quick lookup for every request thereafter.
+// This function will return the correct handler name by building a
+// map of request paths to actual handler names (only during the first
+// request thus providing quick lookup for every request thereafter).
 func (e *EchoWrapper) handlerName(c echo.Context) string {
-	// if first request
-	if e.handlerNames == nil {
+	// only perform once
+	e.once.Do(func() {
 		// build map of request paths to handler names
 		routes := c.Echo().Routes()
 		e.handlerNames = make(map[string]string, len(routes))
 		for _, r := range c.Echo().Routes() {
 			e.handlerNames[r.Method+r.Path] = r.Name
 		}
-	}
+	})
 
 	// lookup handler name for this request
 	return e.handlerNames[c.Request().Method+c.Path()]
--- a/wrappers/hnyecho/echo.go
+++ b/wrappers/hnyecho/echo.go
@@ -41,15 +41,6 @@ func (e *EchoWrapper) Middleware() echo.MiddlewareFunc {
 				span.AddField("app."+name, c.Param(name))
 			}
 
-			// add field for each query param
-			for name, values := range c.QueryParams() {
-				if len(values) == 1 {
-					span.AddField("app."+name, values[0])
-				} else {
-					span.AddField("app."+name, values)
-				}
-			}
-
 			// invoke next middleware in chain
 			err := next(c)
 
--- a/wrappers/hnyecho/echo.go
+++ b/wrappers/hnyecho/echo.go
@@ -1,0 +1,81 @@
+package hnyecho
+
+import (
+	"github.com/jamietsao/beeline-go/wrappers/common"
+	"github.com/labstack/echo"
+)
+
+// EchoWrapper provides Honeycomb instrumentation for the Echo router via middleware
+type (
+	EchoWrapper struct {
+		handlerNames map[string]string
+	}
+)
+
+// New returns a new EchoWrapper struct
+func New() *EchoWrapper {
+	return &EchoWrapper{}
+}
+
+// Middleware returns an echo.MiddlewareFunc to be used with Echo.Use()
+func (e *EchoWrapper) Middleware() echo.MiddlewareFunc {
+	return func(next echo.HandlerFunc) echo.HandlerFunc {
+		return func(c echo.Context) error {
+			r := c.Request()
+			// get a new context with our trace from the request
+			ctx, span := common.StartSpanOrTraceFromHTTP(r)
+			defer span.Send()
+			// push the context with our trace and span on to the request
+			c.SetRequest(r.WithContext(ctx))
+
+			// get name of handler
+			handlerName := e.handlerName(c)
+			if handlerName == "" {
+				handlerName = "handler"
+			}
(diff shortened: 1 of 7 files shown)


nathanleclaire:
Awesome!!

cc @maplebed

wrappers/hnyecho/echo.go
```diff
@@ -0,0 +1,81 @@
...
+			return err
+		}
+	}
+}
+
+// Unfortunately the name of c.Handler() is an anonymous function
+// (https://github.com/labstack/echo/blob/master/echo.go#L487-L494).
+// This function will build a map of request paths to handle names during
+// the first request, thus providing quick lookup for every request thereafter.
+func (e *EchoWrapper) handlerName(c echo.Context) string {
```
jamietsao: As per the comment, the name of the handler for a request isn't readily available so I'm building a map of request paths -> handler names for easy lookup.  Looking at the other example wrappers, the handler name seems important to add as a field to the honeycomb event?

wrappers/hnyecho/echo.go
```diff
@@ -0,0 +1,81 @@
...
+			handlerName := e.handlerName(c)
+			if handlerName == "" {
+				handlerName = "handler"
+			}
+			span.AddField("handler.name", handlerName)
+			span.AddField("name", handlerName)
+
+			// add field for each path param
+			for _, name := range c.ParamNames() {
+				span.AddField("app."+name, c.Param(name))
```
jamietsao: One thing I wasn't sure about was the name of the field for both path params and query params (below).  In some of the example wrappers (e.g. gorilla and goji), analogous fields were named with `gorilla.vars.<var>` (or `goji.pat.<var>`).  I decided to go with `app.<var>` as I didn't want to couple the router implementation to the name.  I believe beeline uses `app.<var>` for any custom fields added so I figured this was a good default.

Thoughts?

wrappers/hnyecho/echo.go
```diff
@@ -0,0 +1,81 @@
+package hnyecho
+
+import (
+	"github.com/jamietsao/beeline-go/wrappers/common"
```
jamietsao: If/when this gets merged, do I need to change the import path to be `honeycombio` instead of `jamietsao`?

nathanleclaire: Yeah - in fact you can probably move the `github.com/jamietsao/beeline-go` fork to your local `$GOPATH/go/src/github.com/honeycombio/beeline-go` and go ahead and do it now

wrappers/hnyecho/echo.go
maplebed: I agree the `gorilla.vars.<var>` and `goji.pat.<var>` is super annoying. I am leaning towards something like `route.vars.<var>` or some variant. I would like to keep the `app.*` namespace exclusively for fields added with `beeline.AddField()` so that it's clear "this stuff came from my manual instrumentation" vs. "this stuff came from a middleware in the beeline that's doing detection magic for me". 
I've been hesitant to change the gorilla and goji pattern because it's always annoying to change people's field names out from under them, and they're in the wild now.  And yet, it keeps coming up as a thing that's annoying people. I think the beeline may be headed for a major version bump in the near future; bringing sanity to middleware function names is a good thing to include in that.

jamietsao: Makes sense.  Given that this Echo middleware will be new, I definitely want to use a forward looking naming convention.  I'm good with `route.vars.<var>` (or `path.vars.<var>`?).  Let me know what you prefer or think the next version will go with.

wrappers/hnyecho/echo.go
```diff
@@ -0,0 +1,81 @@
...
+			}
+
+			// add field for each query param
+			for name, values := range c.QueryParams() {
+				if len(values) == 1 {
+					span.AddField("app."+name, values[0])
+				} else {
+					span.AddField("app."+name, values)
+				}
+			}
```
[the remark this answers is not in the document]
jamietsao: That's a good point.  My thinking around supporting query params automatically is to save the need to manually instrument fields that are useful (e.g. `event_id`), but protecting the schema against the dangerous public is smart and the right thing to do.  My original version actually had config flags for both path and query params, but decided to go with a simple more opinionated approach.  Your experience with query params holds a lot of weight so I'm going to remove support for it completely.

wrappers/hnyecho/echo.go
```diff
@@ -0,0 +1,81 @@
...
+	}
+}
+
+// Unfortunately the name of c.Handler() is an anonymous function
+// (https://github.com/labstack/echo/blob/master/echo.go#L487-L494).
+// This function will build a map of request paths to actual handler names during
+// the first request, thus providing quick lookup for every request thereafter.
+func (e *EchoWrapper) handlerName(c echo.Context) string {
+	// if first request
+	if e.handlerNames == nil {
```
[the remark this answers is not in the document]
jamietsao: Yeah, there's no guarantee that routes won't get added after the call to `Echo.Use()`.  I'll look into this possible race condition and protect against it if needed.   Thanks!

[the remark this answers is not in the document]
jamietsao: I'm using `sync.Once` to prevent a possible race condition

wrappers/hnyecho/echo.go
jamietsao: @maplebed - I implemented all of your suggestions except this one.  Any preference for the prefix here?

wrappers/hnyecho/echo.go
maplebed: 👍

wrappers/hnyecho/echo.go
maplebed: We're having hotly contested conversations about what's the best naming pattern. I think we're leaning towards `route`s rather than `path`s, and are on the fence about `param`s vs. `var`s. Sorry for the delay, but one more day would be lovely to get consensus.
(argument for params is that it echoes query params, which is (so far as I can find) the only one of these that actually has a canonical name. also echoes a function definition, which has parameters (though when called they're arguments?). Serious bikeshed time. One more day, promise.)

jamietsao: Totally understand.  One of the two hardest things in computer science right?

maplebed: Ok. What we decided is that the `request.*` has things that are actually in the request, but that the routing engine, handlers, and variables are not actually part of the request. You can imagine a "has-a" relationship - a request _has a_ URL. A route will handle a request, but it's not the same thing, so routing gets its own top level namespace. A route does _have a_ handler and parameters. 

* `request.url` has the entire URL, unchanged as the user sent it (eg `/a/123?page=2`)
* `request.path` has the path portion of the URL (eg `/a/123`)
* `request.query` has the query portion of the URL (eg `page=2`)
* `route` has the matched portion of the path with variables eg `/a/:id/`
* `route.params.*` has all the variable names eg `route.params.id`
* `route.handler` has the name of the function defined by the route to handle the request

Does that sound reasonable?

jamietsao: @maplebed - Yup, I like it.  I'll go ahead and make these changes.  Thanks!

jamietsao: @maplebed  - I added the `route` fields as per your suggestion.  Should the `route` field for the matched portion of the path be `route.path` (similar to `request.path`)?

wrappers/hnyecho/echo.go
```diff
@@ -0,0 +1,80 @@
...
+			// get name of handler
+			handlerName := e.handlerName(c)
+			if handlerName == "" {
+				handlerName = "handler"
+			}
+			span.AddField("handler.name", handlerName)
+			span.AddField("name", handlerName)
+
+			// add route related fields
+			span.AddField("route", c.Path())
```
jamietsao: @maplebed - As per your suggestion, I added the following `route` fields.  Should this one that contains the matched route be `route.path` instead?

jamietsao:
@maplebed - I've made the route changes and also updated the tests/documentation.  I just had one question (see other comment).  Let me know if there's anything else.

wrappers/hnyecho/echo.go
maplebed: That was one suggestion, but I fear it will lead to more confusion (which path? request path? route path? how are they different again?). I also expect that it's getting into bikeshed territory, so am content to just go with what's here for now.

maplebed:
Rock on. Super thankful that you've written this.

wrappers/hnyecho/echo.go
jamietsao: Works for me!

jamietsao:
@maplebed - Not sure why the Travis build is failing.  The tests all pass when I run them locally.  Any clue?

maplebed:
Update - for me, tests fail when I use the `master` branch of `labstack/echo`. By pinning to that repo to `v4.0.0` tests pass. I'm going to guess that's what's going on; trying to confirm.

jamietsao:
Ah, interesting.  Unless I'm mistaken, I don't see any package manager being used for beeline (e.g. `dep`).  How would one specify a specific version for a dependency?

maplebed:
Although... they fail in a different way than the way the travis-ci build fails. :/ 
but Triumph! They fail for me in the same way as for Travis-CI when i run them on a Linux host instead of my Mac laptop. aaaaand it's not just the echo tests that fail. 🤦‍♂️

jamietsao:
@maplebed - any luck with this?

eanakashima:
@jamietsao apologies for the delay — @maplebed is OOO this week but our on-call engineer is taking a look. Hoping to have an update for you soon.

akvanhar:
@jamietsao , I've been trying to dig into this, and when I clone your branch, the tests fail for me locally. I'm running go 1.12 - something in the middlewear is failing, but we're not sure what yet.

maplebed:
I set up tests on a linux server again today. 
* tests on `master` passed (checking out honeycombio/beeline-go)
* tests on the head of your branch failed (checking out master on jamietsao/beeline-go)

I backed up both my checkout and your clone to the point at which they diverged (commit 0914191ae422d8) and did the same thing and got the same results - worked in the honeycomb checkout, failed in yours.

That's an interesting development - the _same code_ fails in your cloned branch. I tried one of the things you did during development - changing the import statements in your branch to reference your branch instead of the main honeycomb repo.

`~/go/src/github.com/jamietsao/beeline-go$ find * -name '*.go' -exec sed -i "s/\"github.com\/honeycombio\/beeline-go/\"github.com\/jamietsao\/beeline-go/g" {} \;`

_ALL TESTS PASS_

Moving to the tip of master in your repo and making the same change all tests pass once again. Something about referencing the honeycomb-based package in parts of the beeline is causing all the tests to fail. 

At this point, I don't want to block merging based on that, as it's clearly independent from any of your changes. I think the time has come to merge and then watch the tests pass once the code is in the main repo.

maplebed:
Just to follow up, after merging the test still failed. This time though, I could recreate the test failure on my laptop, so debugging was significantly improved. I chased it down to the test initializing `libhoney` instead of the `beeline`. When combined with a recent change to give the beeline its own libhoney client on initialization, this meant that the beeline was actually using a `DiscardSender` rather than the `MockOutput` the tests expected. Changing the test to initialize the beeline with a mock client fixed the issue.

jamietsao:
@maplebed - Awesome, this is great news.  Thanks for your help in getting this over the line!  Much appreciated.

PR #53, opened 2019-03-15
Allow override of libhoney client

tredman:
We introduced client objects in https://github.com/honeycombio/libhoney-go/pull/40 - now we want to enable the beeline to use the non-default libhoney client and enable better customization of the client.

This required some refactoring, as the `beeline` module and the `trace` module both shared the `libhoney` global state, and don't share any state within the library. To make this work, I pulled client initialization into its own module, and exposed the client methods as module methods. Since libhoney's `NewClient` returns an error, but beeline's `Init` does not, we don't have a great way to handle failed initializations, so the wrapper methods in the `client` module protect agains the case that client initialization fails and returns us a nil client.


client/client.go
```diff
@@ -0,0 +1,59 @@
...
+	if client != nil {
+		client.AddField(name, val)
+	}
+}
+
+func NewBuilder() *libhoney.Builder {
+	if client != nil {
+		return client.NewBuilder()
+	}
+	return &libhoney.Builder{}
```
maplebed: Most of the rest noop if the client is nil; how come this one returns a builder? And if it does return a libhoney builder, does that mean that `Close` should call `libhoney.Close`, and same with add field, flush, etc.? I see that the builder is the only one with a return value and it's rude to return `nil`, but it feels a bit unexpectedly inconsistent. hrmph.

maplebed: I see though, where trace needs a builder...  ok. Good for now.

PR #54, opened 2019-03-21
ev can't have been already sampled here so multiplying is incorrect

maplebed:
The multiplication of sample rate here is incorrect. It came from an assumption that `ev` could have already been sampled, but the code makes using a sampler hook an _alternative_ to the global sample rate, not in addition to it. Since sampling only happens once, the rates should not be multiplied.


CHANGELOG.md
```diff
@@ -1,11 +1,58 @@
+Release v0.4.1 (2019-03-21)
+==
+
+### Bugfixes
```
tredman: yay release notes!

PR #57, opened 2019-04-08
Fix data race for children spans.

carlosgaldino:
This patch makes sure that before checking how many children a span has; when calling `send()`; we will hold the `childrenLock` to avoid a data race if another goroutine is at the same time creating a new span and writing to `s.children`.

The lock is held here: https://github.com/honeycombio/beeline-go/blob/015109f8347c1922dcc0cc3b96d706c24fd96ef8/trace/trace.go#L438-L440

But when checking how many items the slice contains the lock is not held: https://github.com/honeycombio/beeline-go/blob/015109f8347c1922dcc0cc3b96d706c24fd96ef8/trace/trace.go#L389

### DIFF
--- a/trace/trace.go
+++ b/trace/trace.go
@@ -375,6 +375,7 @@ func (s *Span) send() {
 		s.AddField(k, v)
 	}
 
+	s.childrenLock.Lock()
 	// classify span type
 	var spanType string
 	switch {
@@ -391,6 +392,7 @@ func (s *Span) send() {
 	default:
 		spanType = "mid"
 	}
+	s.childrenLock.Unlock()
 	s.AddField("meta.span_type", spanType)
 
 	if spanType == "root" {
--- a/trace/trace_test.go
+++ b/trace/trace_test.go
@@ -264,6 +264,29 @@ func TestCreateAsyncSpanDoesNotCauseRaceInSend(t *testing.T) {
 	wg.Wait()
 }
 
+// TestCreateSubSpanDoesNotCauseRaceInSend exists because when sending a root
+// span, you don't consider the length of the children to determine if the span
+// is a leaf. By doing the same test as above on a subspan, we test for a race
+// on the number of chilrden in a span.
+func TestCreateSubSpanDoesNotCauseRaceInSend(t *testing.T) {
+	setupLibhoney()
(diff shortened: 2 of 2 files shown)


carlosgaldino:
Let me know if you need more information or if anything needs to be changed. Thanks!

maplebed:
Thank you! 

I was confused why this escaped the race detector, since there is a test that tries to send and add children to a span simultaneously.  Turns out if you are the root span, you never run the `len()` that causes the race because you exit the `switch` statement having found out that you're the root.

Here is a test that triggers the race (along with a sync span test to match the existing async span one); could you add it to the PR?

```
// TestCreateSubSpanDoesNotCauseRaceInSend exists because when sending a root
// span, you don't consider the length of the children to determine if the span
// is a leaf. By doing the same test as above on a subspan, we test for a race
// on the number of chilrden in a span
func TestCreateSubSpanDoesNotCauseRaceInSend(t *testing.T) {
	setupLibhoney()
	ctx, tr := NewTrace(context.Background(), t.Name())
	rs := tr.GetRootSpan()
	ctx, subsp := rs.CreateChild(ctx)

	wg := &sync.WaitGroup{}
	wg.Add(2)
	go func() {
		subsp.Send()
		wg.Done()
	}()
	go func() {
		subsp.CreateChild(ctx)
		wg.Done()
	}()
	wg.Wait()
}

func TestChildAndParentSendsDoNotRace(t *testing.T) {
	setupLibhoney()

	wg := &sync.WaitGroup{}
	wg.Add(10)
	for i := 0; i < 5; i++ {
		ctx, tr := NewTrace(context.Background(), t.Name())
		rs := tr.GetRootSpan()

		go func() {
			rs.Send()
			wg.Done()
		}()
		go func() {
			ct, sp1 := rs.CreateChild(ctx)
			ct, sp2 := sp1.CreateChild(ct)
			ct, sp3 := sp2.CreateChild(ct)
			sp1.Send()
			sp2.Send()
			sp3.Send()

			wg.Done()
		}()
	}

	wg.Wait()
}
```

carlosgaldino:
Thanks for the reply @maplebed. I pushed the changes you mentioned. Cheers!

PR #73, opened 2019-09-16
hnysqlx: Provide a method for accessing the inner DB

alex:
This can be useful when you want to use APIs that are not a part of the standard DB API.

In my specific case I want to use `pgx`'s `COPY FROM` support, which requires me to access the underlying `pgx` conn.


alex:
thanks!

PR #74, opened 2019-09-17
defer a function so we can close over err instead of passing it as an arg

toshok:
When we pass it as an arg, it's evaluated at `defer`-time.  This means we'll never send a non-nil error.

Closing over it instead should allow it to be updated and we'll send the right thing.


maplebed:
Nice catch. With this change `err` will be evaluated at the deferred time instead of on the line it's declared.

toshok:
build is red due to go 1.9 failing because of a dependency.  going to land anyway :yolo:

Issue #69, opened 2019-07-10
Echo middleware incompatible with v4


noamt:
The latest version of Echo [requires](https://github.com/labstack/echo#supported-go-versions) the you explicitly import the module using the `/v4` prefix:

> Any of these versions will allow you to import Echo as github.com/labstack/echo/v4 which is the recommended way of using Echo going forward.

So the current Echo middleware is usable only with version 3:
```go
./main.go:56:33: cannot use hnyecho.New().Middleware() (type "github.com/labstack/echo".MiddlewareFunc) as type "github.com/labstack/echo/v4".MiddlewareFunc in argument to e.Use
```

 In the meantime I'm able to work around this issue by duplicating the code from the middleware into my own middleware that imports `/v4`

wolfeidau:
This looks to be related to the gomod support, it would be good to get this fixed as most new tools and libraries use echo v4.

vreynolds:
Closed by #99

PR #77, opened 2019-09-18
#69 - Echo middleware incompatible with v4

noamt:
Update to the latest release of Echo

### DIFF
--- a/wrappers/hnyecho/echo.go
+++ b/wrappers/hnyecho/echo.go
@@ -4,7 +4,7 @@ import (
 	"sync"
 
 	"github.com/honeycombio/beeline-go/wrappers/common"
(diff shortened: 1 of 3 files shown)


maplebed:
Thanks! Sorry for the delay.

maplebed:
I'm sorry, I reverted the commit. The CircleCI tests (and another try on my local machine) failed with `package github.com/labstack/echo/v4: cannot find package "github.com/labstack/echo/v4"`. I expect the fix is simple, but in order to keep master green I did a revert. I will happily look at another attempt to use the upgraded version of `echo`.

noamt:
I suspect that is because the project doesn't use the go modules of 1.11
Would you consider a PR that adds module support?

PR #82, opened 2019-10-17
First pass at go database stats

dsoo:
Add database stats from sql.DBStats to database calls to help understand database state.

### DIFF
--- a/wrappers/common/common_go1.11.go
+++ b/wrappers/common/common_go1.11.go
@@ -1,0 +1,26 @@
+//+build go1.11
+
+package common
+
+import (
+	"database/sql"
+
+	"github.com/honeycombio/beeline-go/trace"
+	libhoney "github.com/honeycombio/libhoney-go"
+)
+
+func addDBStatsToEvent(ev *libhoney.Event, stats sql.DBStats) {
+	ev.AddField("db.open_conns", stats.OpenConnections)
+	ev.AddField("db.conns_in_use", stats.InUse)
+	ev.AddField("db.conns_idle", stats.Idle)
+	ev.AddField("db.wait_count", stats.WaitCount)
+	ev.AddField("db.wait_duration", stats.WaitDuration)
+}
+
+func addDBStatsToSpan(span *trace.Span, stats sql.DBStats) {
+	span.AddField("db.open_conns", stats.OpenConnections)
+	span.AddField("db.conns_in_use", stats.InUse)
+	span.AddField("db.conns_idle", stats.Idle)
+	span.AddField("db.wait_count", stats.WaitCount)
+	span.AddField("db.wait_duration", stats.WaitDuration)
+}
--- a/wrappers/hnysqlx/sqlx.go
+++ b/wrappers/hnysqlx/sqlx.go
@@ -1456,7 +1456,7 @@ func (s *Stmt) Unsafe() *Stmt {
 
 func (s *Stmt) Close() error {
 	var err error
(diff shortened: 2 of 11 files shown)


dsoo:
This is the first pass at implementing these - I don't know that much about our go beelines so lots of design input is welcome.

wrappers/hnysqlx/sqlx.go
```diff
@@ -898,7 +904,7 @@ func (db *DB) SetMaxOpenConns(n int) {
 
 func (db *DB) Stats() sql.DBStats {
 	var err error
-	_, sender := common.BuildDBEvent(db.Builder, "")
+	_, sender := common.BuildDBEvent(db.Builder, db.Stats(), "")
```
maplebed: This looks to me like an endless loop.

wrappers/common/common_go1.11.go
```diff
@@ -0,0 +1,26 @@
...
+	"github.com/honeycombio/beeline-go/trace"
+	libhoney "github.com/honeycombio/libhoney-go"
+)
+
+func addDBStatsToEvent(ev *libhoney.Event, stats sql.DBStats) {
+	ev.AddField("db.open_connections", stats.OpenConnections)
+	ev.AddField("db.connections_in_use", stats.InUse)
+	ev.AddField("db.connections_idle", stats.Idle)
+	ev.AddField("db.wait_count", stats.WaitCount)
+	ev.AddField("db.wait_duration", stats.WaitDuration)
```
maplebed: The godoc makes me think these (`WaitCount` and `WaitDuration`)are both an ever-increasing counters

> The total number of connections waited for
> The total time blocked waiting for a new connection

Is that total ... over the lifetime of the process? If so, I would leave it out.

dsoo: Yeah, I was thinking about those - they ARE over lifetime of the process, but they are super important signals and give you really good signal as to how much impact database contention is having during an incident. In normal operation for most systems, these values should be zero. For example, we could use this to trivially determine if we are running into DB contention issues due to cache filling when shepherd starts, which would be very difficult to determine using the other stats.

wrappers/hnysqlx/sqlx.go
dsoo: Ah, whoops, good catch. Yeah, we don't need to add this here.

dsoo:
Looks like dbstats is cheap - it has a mutex, but the operations within the mutex are constant time.

https://golang.org/src/database/sql/sql.go?s=10849:12619#L1007

wrappers/hnysqlx/sqlx.go
dsoo: Do you think we even should be creating an event for this call? It can't even return an error, it's super fast, and if I'm adding it to every database call that will immediately double the number of database-related events emitted per DB call.

dsoo:
@maplebed - fixed various things, including adding some super-basic tests. Also did some testing on a host in dogfood, as far as I can tell nothing went horribly wrong with the host, and we are still performing okay with the additional instrumentation.
https://ui-kibble.honeycomb.io/dogfood/datasets/shepherd/result/t1LDprJ79Kj

wrappers/hnysqlx/sqlx.go
maplebed: I am on board to skip making a span for this call. Doubling the number of spans just to say "no really, every DB call also calls `Stats`" seems pointless.

PR #88, opened 2019-12-23
[rfc] add callback to delegate determining trace headers

maplebed:
There are two open questions that give rise to this PR as one possible solution:

* how do we configure the beeline to selectively trust incoming trace headers? There are times when an application may handle traffic from trusted and untrusted parties and only the trusted partiies should be allowed to hand the beeline trace headers that it will obey
* how do we enable better cross-trace-product compatibility? It would be nice to listen for AWS ALB tracing headers instead of the beeline headers (so that an AWS load balancer may start the trace) or have a beeline-instrumented process downstream from an open tracing instrumented process and so on. 

This change enables both of these situations to be handled by delegating the code that looks for incoming headers to the caller. The beeline doesn't want to try and understand request authorization (whether or not to trust upstream headers) - that code is better managed by the application. The beeline wants enough flexibility to be easily started downstream of any process, whether or not it comes via HTTP. Accepting a propagation object instead of serialized headers increases the flexibility of callers when starting a trace by pulling something from a queue or other inter-process communication protocols.


MikeGoldsmith:
We’re going to close this for now as there as been no recent activity. Feel free to re-open if it’s still relevant.

maplebed:
more than just abandoned, this PR is made obsolete by https://github.com/honeycombio/beeline-go/pull/118, a more complete implementation of the same idea.

Issue #51, opened 2019-02-21
Initializing the beeline with an empty writekey causes it to send events with a default


maplebed:
If you initialize a `libhoney` with an empty API key, it refuses to send any events, effectively turning the instrumentation into a noop. If you do the same with the beeline, it sends events to Honeycomb with a default API key. This is unexpected. Initializing the beeline with an empty API key should disable sending events to Honeycomb.

vreynolds:
Hello,

We will be closing this issue as it is a low priority for us. It is unlikely that we'll ever get to it, and so we'd like to set expectations accordingly.

As we enter 2022 Q1, we are trimming our OSS backlog. This is so that we can focus better on areas that are more aligned with the OpenTelemetry-focused direction of telemetry ingest for Honeycomb.

If this issue is important to you, please feel free to ping here and we can discuss/re-open.

PR #89, opened 2019-12-24
Accept an empty WriteKey to disable Honeycomb

jdub:
Fixes honeycombio/beeline-go#51.


MikeGoldsmith:
We’re going to close this for now. It would require a major bump due to a significant change in default behaviour.

Feel free to re-open, if it’s still relevant.

PR #97, opened 2020-03-03
Add ability to extract arbitrary request headers as properties

memory:
- add the referer header (if present) by default
- allow setting ExtractCustomHeaders to a list of header names

### DIFF
--- a/wrappers/common/common.go
+++ b/wrappers/common/common.go
@@ -14,6 +14,10 @@ import (
 	libhoney "github.com/honeycombio/libhoney-go"
 )
 
+var (
+	ExtractCustomHeaders []string
+)
+
 type ResponseWriter struct {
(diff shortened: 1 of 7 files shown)


memory:
Note: I haven't updated the docs yet. Happy to add that in when there's consensus about the approach.

wrappers/common/common.go
```diff
@@ -14,6 +14,10 @@ import (
 	libhoney "github.com/honeycombio/libhoney-go"
 )
 
+var (
+	ExtractCustomHeaders []string
```
seh: Please don't introduce global state like this. You could instead introduce a new function that accepts this string slice as a parameter in addition to a `*http.Request`, pass the latter to the existing `GetRequestProps` function, capture the returned map, and add to that map by consulting your string slice before returning it—all in your own code. There's no need to touch `GetRequestProps` to do this.

PR #99, opened 2020-03-05
Use go1.11 modules instead of older modules system.

lizthegrey:
Supersedes https://github.com/honeycombio/beeline-go/pull/98 by @memory


.circleci/config.yml
```diff
@@ -6,32 +6,36 @@ executors:
...
+        default: "12"
     executor:
       name: go
       goversion: "<< parameters.goversion >>"
     steps:
       - checkout
-      - run: go get -v -t -d ./...
-      - run: go test -race -v ./...
+      - run:
+          name: go-get
```
memory: I think if you want to ensure in the build that PRs include any necessary updates to go.mod/go.sum, you'll want to run `go vet -mod=readonly ./...` before this?  (Or just skip the go-get stage and run `go test -mod=readonly -race -v ./...` as the single stage? I'll confess I'm not 100% clear on when `go get` does and does not happen transitively.)

PR #108, opened 2020-05-08
Provide hooks and utils for trace header interop.

paulosman:
This change introduces HTTPPropagators to the Go beeline. Propagators are used for extracting trace context propagation headers from incoming HTTP requests and injecting them into outgoing HTTP requests. 

We include propagators for Honeycomb and Amazon X-Ray trace header formats. We also wrap the OpenTelemetry Go SDK default propagator so that we can parse and generate W3C Trace Context headers.

The interface is extensible as well. Additional header formats can be supported with the addition of HTTPPropagator implementations and custom formats can be plugged in by users of the beeline.

Sample client code using the hooks:

```
package main

import (
        beeline "github.com/honeycombio/beeline-go"
        "github.com/honeycombio/beeline-go/trace"
        "github.com/honeycombio/beeline-go/propagation"
)

// An example that parses Honeycomb trace context headers and
// instruments outgoing HTTP calls with Honeycomb and W3C trace context
// headers.
func withPropagators(ctx context.Context, r *http.Request) []trace.HTTPPropagator {
	return []trace.HTTPPropagator{
		propagation.HoneycombHTTPPropagator{},
		propagation.W3CHTTPPropagator{},
	}
}

func withParsers(ctx context.Context, r *http.Request) []trace.HTTPPropagator {
	return []trace.HTTPPropagator{
                propagation.HoneycombHTTPPropagator{},
        }
}

func main() {
	beeline.Init(beeline.Config{
		WriteKey: "Your Write Key",
		Dataset: "distributed-app",
		TraceHTTPHeaderParserHook: withParsers,
		TraceHTTPHeaderPropagationHook: withPropagators,
	})
	defer beeline.Close()
	...
}
```

The interface for Propagators is:

```
// HeaderSupplier is an interface that specifies methods for storing and retrieving
// key value pairs. It is implemented by http.Headers.
type HeaderSupplier interface {
        Set(key string, value string)
        Get(key string) string
}

// HTTPPropagator wraps the interface with the same name from the OpenTelemetry Go SDK.
// It intentionally uses different method names to highlight the fact that it's API is
// different than the analagous object in the OpenTelemetry Go SDK.
type HTTPPropagator interface {
        // Parse is responsible for retrieving the appropriate header(s) from header and
        // using the information contained within to create and return a SpanContext object.
        Parse(ctx context.Context, header HeaderSupplier) *SpanContext
        // Insert is responsible for retrieving the appropriate information about the current
        // span from ctx and using the information contained within to generate a serialized
        // header(s) which are then inserted into header.
        Insert(ctx context.Context, header HeaderSupplier)
}
```


trace/trace.go
```diff
@@ -2,25 +2,44 @@ package trace
...
 	"github.com/honeycombio/beeline-go/sample"
 	libhoney "github.com/honeycombio/libhoney-go"
+	otelprop "go.opentelemetry.io/otel/api/propagation"
 )
 
 var GlobalConfig Config
 
+// HTTPPropagator wraps the interface with the same name from the OpenTelemetry Go SDK.
+type HTTPPropagator interface {
+	otelprop.HTTPPropagator
```
maplebed, line 24: All the details of interfaces fulfillment are a little fuzzy but I think it's a little dangerous to embed their propagator directly. The effect is that if they add a function to the interface our implementations no longer meet our interface and so immediately fail to compile. This makes revving versions difficult. It would be better to call out the specific functions that we need our interfaces to provide and then we can lag behind their version bumps without compatibility issues.

trace/otel_span.go
```diff
@@ -0,0 +1,110 @@
...
+	
+	if sp == nil {
+		return otelSpan{}
+	}
+
+	// OpenTelemetry uses 16 byte trace ids and 8 byte span ids. Because honeycomb
+	// uses uuids for both, we have to truncate span ids before sending them on in
+	// trace headers.
+	traceId, _ := core.TraceIDFromHex(strings.ReplaceAll(sp.GetTrace().GetTraceID(), "-", ""))
+	spanId, _ := core.SpanIDFromHex(strings.ReplaceAll(sp.spanID, "-", "")[:16])
```
paulosman, line 81: This will need to be thought out more. It's necessary when sending UUID like strings we generate to Tel, but I'm not comfortable with just an arbitrary chop.

trace/otel_span.go
```diff
@@ -0,0 +1,110 @@
...
+// OCSpanContextToHCSpanContext converts an OpenTelemetry Go SDK core.SpanContext
+// object into a Honeycomb SpanContext. Note that Honeycomb trace IDs are by default
+// generated as strings in UUID format. We have to reformat OTel trace ids in a similar
+// way for propagation to work.
+func OCSpanContextToHCSpanContext(spanContext core.SpanContext) *SpanContext {
+	traceID := hex.EncodeToString(spanContext.TraceID[:])
+	if len(traceID) == 32 {
+		traceID = fmt.Sprintf(
+			"%s-%s-%s-%s-%s",
+			traceID[0:8], traceID[8:12], traceID[12:16], traceID[16:20], traceID[20:])
```
paulosman, line 102: I'm not happy with this, but needed to reformat trace ids coming in from W3C transparent headers in order for them to match those generated by beelines.

trace/otel_span.go
```diff
@@ -0,0 +1,110 @@
...
+
+	if sp == nil {
+		return otelSpan{}
+	}
+
+	// OpenTelemetry uses 16 byte trace ids and 8 byte span ids. Because honeycomb
+	// uses uuids for both, we have to truncate span ids before sending them on in
+	// trace headers.
+	traceID, _ := core.TraceIDFromHex(strings.ReplaceAll(sp.GetTrace().GetTraceID(), "-", ""))
+	spanID, _ := core.SpanIDFromHex(strings.ReplaceAll(sp.spanID, "-", "")[:16])
```
paulosman, line 81: I'd like to think through the implications of this a bit before landing. Feels wrong to just chop off half of the span id, but OTel spans are 8 bytes.

trace/trace.go
paulosman: Good call. Replaced with a list of the methods we need to implement in our propagators.

paulosman:
This is just an "FYI" for now. I plan to revise some parts of this, especially the trace ID formatting/chopping. Also going to add tests and some helper methods for common use cases.

toshok:
re: the asymmetry I mentioned in slack, I think I'd much prefer if we have slices for both parser and propagation hooks.  This way we don't have a declarative list in one place and imperative code in another.

e.g. this:
```
func withPropagators(ctx context.Context, r *http.Request) []trace.HTTPPropagator {
        // propagators emit all headers listed in the slice
	return []trace.HTTPPropagator{
		propagation.HoneycombHTTPPropagator(),
		oteltrace.DefaultHTTPPropagator(),
	}
}

func parseHeaders(ctx context.Context, r *http.Request) []trace.HTTPPropagator {
        // parser behavior is "first parser that returns a parsed trace context we use".  
        // if you only have one parser, return the slice with a single element.  If you need to
        // support multiple clients using different instrumentation, add them here.
	return []trace.HTTPPropagator{
		propagation.HoneycombHTTPPropagator(),
		oteltrace.DefaultHTTPPropagator(),
	}
}

func main() {
	beeline.Init(beeline.Config{
		WriteKey: "Your Write Key",
		Dataset: "distributed-app",
		TraceHTTPHeaderParserHook: parseHeaders,
		TraceHTTPHeaderPropagationHook: withPropagators,
	})
	defer beeline.Close()
	...
}
```

This lets services/teams on either side of the communication channel change their behavior more independently of what's on the other side.

toshok:
similar to what otel-go does too, e.g. https://github.com/open-telemetry/opentelemetry-go/blob/587cde3352d785e7107b70166993c76dd709e2fa/plugin/httptrace/httptrace.go#L35

returning the propagator slice is analogous to `global.Propagators()`.

otel treats extractors the same as injectors (our propagators) in that it runs all of them.  If we're assuming that all the data will be the same for the fields we care about, we could do that as well.

paulosman:
> similar to what otel-go does too, e.g. https://github.com/open-telemetry/opentelemetry-go/blob/587cde3352d785e7107b70166993c76dd709e2fa/plugin/httptrace/httptrace.go#L35
> 
> returning the propagator slice is analogous to `global.Propagators()`.
> 
> otel treats extractors the same as injectors (our propagators) in that it runs all of them. If we're assuming that all the data will be the same for the fields we care about, we could do that as well.

I think I'm on board with this. Basically, we're supporting the following use case as typical:

- Incoming services could have either W3C headers or Honeycomb headers (for example). Use both propagators when parsing incoming requests. In the rare case that a service has both wire formats, the data should be the same anyway. 

Is that an accurate restating? I do very much like the symmetry.

beeline.go
```diff
@@ -52,7 +53,20 @@ type Config struct {
 	// event before it gets sent to Honeycomb. Does not get invoked if the event
 	// is going to be dropped because of sampling. Runs after the SamplerHook.
 	PresendHook func(map[string]interface{})
-
+	// TraceHeaderParserHook is a function call that will get run with the context
+	// and HTTP request on all incoming requests. The function registered may
```
seh, line 57: Does "may" here mean that this function can return nil as its first return value?

beeline.go
```diff
@@ -52,7 +53,20 @@ type Config struct {
...
+	// construct and return a SpanContext, which will be used in the subsequent
+	// creation of trace and span objects. This is useful when parsing trace
+	// propagation headers from upstream services. HTTPPropagator objects defined
+	// in the propagation package or elsewhere may be used to do the parsing.
+	TraceHTTPHeaderParserHook func(ctx context.Context, r *http.Request) (*trace.SpanContext, context.Context)
+	// TraceHTTPHeaderPropagationHook is a function call that will get run with
+	// the context and HTTP request on all outgoing http requests. The function
```
seh, line 64:
```suggestion
	// the context and HTTP request on all outgoing HTTP requests. The function
```

beeline.go
```diff
@@ -52,7 +53,20 @@ type Config struct {
...
 	PresendHook func(map[string]interface{})
-
+	// TraceHeaderParserHook is a function call that will get run with the context
+	// and HTTP request on all incoming requests. The function registered may
+	// construct and return a SpanContext, which will be used in the subsequent
+	// creation of trace and span objects. This is useful when parsing trace
+	// propagation headers from upstream services. HTTPPropagator objects defined
+	// in the propagation package or elsewhere may be used to do the parsing.
+	TraceHTTPHeaderParserHook func(ctx context.Context, r *http.Request) (*trace.SpanContext, context.Context)
+	// TraceHTTPHeaderPropagationHook is a function call that will get run with
```
seh, line 63:
```suggestion
	// TraceHTTPHeaderPropagationHook is a function that will be invoked with
```

beeline.go
```diff
@@ -179,6 +193,15 @@ func Init(config Config) {
 	if config.PresendHook != nil {
 		trace.GlobalConfig.PresendHook = config.PresendHook
 	}
+
+	if config.TraceHTTPHeaderParserHook != nil {
```
seh, line 197:
```suggestion
	if len(config.TraceHTTPHeaderParserHook) > 0 {
```

beeline.go
```diff
@@ -179,6 +193,15 @@ func Init(config Config) {
...
+		trace.GlobalConfig.TraceHTTPHeaderParserHook = config.TraceHTTPHeaderParserHook
+	}
+
+	if config.TraceHTTPHeaderPropagationHook != nil {
```
seh, line 201:
```suggestion
	if len(config.TraceHTTPHeaderPropagationHook) > 0 {
```

propagation/propagation.go
```diff
@@ -1,13 +1,35 @@
...
+	otelprop "go.opentelemetry.io/otel/api/propagation"
+)
+
+const (
+	HoneycombTracePropagationHTTPHeader = "X-Honeycomb-Trace"
+	HoneycombTracePropagationVersion    = 1
+	AmazonTracePropagationHTTPHeader    = "X-Amzn-Trace-Id"
 )
 
+type PropagationError struct {
```
seh, line 21: Why export this type?

propagation/propagation.go
```diff
@@ -1,13 +1,35 @@
...
 	"fmt"
 	"net/url"
 	"strings"
+
+	"github.com/honeycombio/beeline-go/trace"
+	otelprop "go.opentelemetry.io/otel/api/propagation"
+)
+
+const (
+	HoneycombTracePropagationHTTPHeader = "X-Honeycomb-Trace"
```
seh, line 16: Why export these constants?

propagation/propagation.go
```diff
@@ -23,97 +45,159 @@ import (
...
-	Dataset      string
-	TraceContext map[string]interface{}
-}
-
-type PropagationError struct {
-	message      string
-	wrappedError error
+// HoneycombHTTPPropagator returns an HTTPPropagator that understands Honeycomb
+// trace context headers.
+func HoneycombHTTPPropagator() trace.HTTPPropagator {
```
seh, line 52: I vote for either not exporting `HoneycombContext` **or** returning `HoneycombContext` here. Don't return the interface type unless you're trying to hide the concrete type.

propagation/propagation.go
```diff
@@ -23,97 +45,159 @@ import (
...
 }
 
-func UnmarshalTraceContext(header string) (*Propagation, error) {
-	// pull the version out of the header
-	getVer := strings.SplitN(header, ";", 2)
-	if getVer[0] == "1" {
-		return UnmarshalTraceContextV1(getVer[1])
+// Unmarshal is a helper method that is not part of the HTTPPropagator
+// interface. It accepts a context and a supplier and returns a SpanContext. It
+// is assumed that supplier contains a valid honeycomb trace context header.
```
seh, line 82:
```suggestion
// is assumed that supplier contains a valid Honeycomb trace context header.
```

propagation/propagation.go
```diff
@@ -23,97 +45,159 @@ import (
...
-		TracePropagationVersion,
-		prop.TraceID,
-		prop.ParentID,
-		datasetClause,
-		tcB64,
-	)
+// Marshal is a helper method that is not part of the HTTPPropagator interface.
+// It puts the supplied SpanContext into the context, injects the header into
+// the supplier and then returns the resulting header string as well as the
+// context. This is useful for testing.
```
seh, line 73: Does it need to be exported?

propagation/propagation.go
```diff
@@ -23,97 +45,159 @@ import (
...
 
-func UnmarshalTraceContext(header string) (*Propagation, error) {
-	// pull the version out of the header
-	getVer := strings.SplitN(header, ";", 2)
-	if getVer[0] == "1" {
-		return UnmarshalTraceContextV1(getVer[1])
+// Unmarshal is a helper method that is not part of the HTTPPropagator
+// interface. It accepts a context and a supplier and returns a SpanContext. It
+// is assumed that supplier contains a valid honeycomb trace context header.
+// This is useful for testing.
```
seh, line 83: Does it need to be exported?

propagation/propagation.go
```diff
@@ -23,97 +45,159 @@ import (
...
+func (HoneycombContext) GetAllKeys() []string {
+	return []string{HoneycombTracePropagationHTTPHeader}
+}
+
+// AmazonContext understands how to parse and generate Amazon trace propagation headers
+type AmazonContext struct{}
+
+// AmazonHTTPPropagator returns an HTTPPropagator that understands Amazon trace context
+// headers.
+func AmazonHTTPPropagator() trace.HTTPPropagator {
```
seh, line 151: Like `HoneycombContext` above, I recommend either returning type `AmazonContext` **or** unexporting the type.

trace/otel_span.go
```diff
@@ -0,0 +1,103 @@
...
+	"encoding/hex"
+	"strings"
+	"time"
+
+	"go.opentelemetry.io/otel/api/core"
+	oteltrace "go.opentelemetry.io/otel/api/trace"
+	"google.golang.org/grpc/codes"
+)
+
+// implements otel/api/trace.Span. Has methods defined in that interface, but
```
seh, line 14:
```suggestion
// otelSpan implements otel/api/trace.Span. Though it defines methods to satisfy that interface,
```

trace/otel_span.go
```diff
@@ -0,0 +1,103 @@
...
+	}
+
+	return otelSpan{
+		spanContext: spanContext,
+	}
+}
+
+// OCSpanContextToHCSpanContext converts an OpenTelemetry Go SDK core.SpanContext
+// object into a Honeycomb SpanContext. Note that Honeycomb trace IDs are by default
+// generated as strings in UUID format. We have to reformat OTel trace ids in a similar
```
seh, line 94:
```suggestion
// generated as strings in UUID format. We have to reformat OTel trace IDs in a similar
```

trace/trace.go
```diff
@@ -2,25 +2,48 @@ package trace
...
 	// PresendHook is a function to mutate spans just before they are sent to
 	// Honeycomb. See the docs for `beeline.Config` for a full description.
 	PresendHook func(map[string]interface{})
+	// TraceHTTPHeaderParserHook is a function to parse incoming trace context
+	// headers and turn them into span context. See the docs for `beeline.Config`
+	// for a full description.
+	TraceHTTPHeaderParserHook func(ctx context.Context, r *http.Request) (*SpanContext, context.Context)
+	// TraceHTTPHeaderPropagationHook is a function that allows a user of the beeline
+	// to specify which HTTPPropagators should be used when propagating trace context
+	// via HTTP headers in outgoing calls. See the docs for `beeline.Config` for a full
```
seh, line 44:
```suggestion
	// via HTTP headers in outgoing calls. See beeline.Config for a full
```

paulosman:
> I didn't review the tests; I was mostly interested in the exported interface.

No tests yet, want to make sure we get the interface right first, so this is perfect. Thank you! Pushed a change to stop unnecessarily exporting types/methods. Will address the other comments today.

beeline.go
paulosman: Updated interface and language to avoid ambiguity. Let me know if the new one is still unclear?

beeline.go
paulosman: :+1:

beeline.go
paulosman: Not sure I understand this suggestion. `config.TraceHTTPHeaderParserHook` is a function, so it's zero-value is `nil`, so this is the correct check, right?

beeline.go
paulosman: Same as above.

propagation/propagation.go
paulosman: Lowercased. There are some other types and functions that might be exported and shouldn't. Will do a full audit before promoting this PR to ready to merge.

propagation/propagation.go
paulosman: I agree. Opted to return `HoneycombContext` (and `AmazonContext` below).

beeline.go
seh: Oh, I'm sorry. I mistook these two fields for the slices I saw elsewhere, thinking that you were checking if a slice field was nil. I rescind my suggestion here and below.

beeline.go
seh: Rescinded, per my comment above.

beeline.go
seh: Looks good. Thank you.

beeline.go
paulosman: I like this suggestion - changed throughout (there were a few other cases of "get run with").

paulosman:
> This is just an "FYI" for now. I plan to revise some parts of this, especially the trace ID formatting/chopping. Also going to add tests and some helper methods for common use cases.

I think this is getting close to ready for review. Still needs tests.

propagation/propagation.go
```diff
@@ -23,97 +45,159 @@ import (
...
+// provided context.
+func (AmazonContext) Extract(ctx context.Context, supplier otelprop.HTTPSupplier) context.Context {
+	header := supplier.Get(amazonTracePropagationHTTPHeader)
+	segments := strings.Split(header, ";")
+
+	sc := &trace.SpanContext{}
+	for _, segment := range segments {
+		keyval := strings.SplitN(segment, "=", 2)
+		switch strings.ToLower(keyval[0]) {
+		case "self":
```
dsoo, line 165: Doesn't the Amazon header use `Root` and `Parent`, as in the `Extract` code below? Should these be defined constants?

trace/otel_span.go
```diff
@@ -0,0 +1,110 @@
...
+
+func (os otelSpan) SetName(name string) {
+	return
+}
+
+// oTelRemoteSpanFromContext looks for an Otel SpanContext in the provided context.
+// If it finds a non-empty context, it returns it, otherwise it returns nil.
+func oTelRemoteSpanFromContext(ctx context.Context) *oteltrace.SpanContext {
+	spanContext := oteltrace.RemoteSpanContextFromContext(ctx)
+	emptyContext := oteltrace.SpanContext{}
```
dsoo, line 68: Is doing this on line line, `if oteltrace.SpanContext{} == spanContext` a more normal way of checking for a zero value struct in Go?

trace/otel_span.go
```diff
@@ -0,0 +1,110 @@
...
+		TraceID: traceID,
+		SpanID:  spanID,
+	}
+
+	return otelSpan{
+		spanContext: spanContext,
+	}
+}
+
+// OCSpanContextToHCSpanContext converts an OpenTelemetry Go SDK core.SpanContext
```
dsoo, line 99: Is this comment still valid? We're no longer doing this, right?

trace/trace.go
```diff
@@ -42,35 +69,44 @@ type Trace struct {
 // NewTrace creates a brand new trace. serializedHeaders is optional, and if
```
dsoo, line 69: Args have changed, update the comment?

trace/trace.go
```diff
@@ -472,3 +577,11 @@ func (s *Span) createChildSpan(ctx context.Context, async bool) (context.Context
 	ctx = PutSpanInContext(ctx, newSpan)
 	return ctx, newSpan
 }
+
+// generateID generates random bytes and encodes them as a hex string. The
+// length of the id depends on the length of the byte slice provided. If
+// io.ReadAll fails, the result will be all zeroes.
```
dsoo, line 583: `io.ReadFull`, not `io.ReadAll` - also, is all zeros the best we can do on failure - not that this should ever happen - although running out of entropy is a thing? I suppose there aren't a lot of better alternatives - although maybe even an incrementing counter might be better than all the IDs being zero.
Also, in researching, you can just use `rand.Read` here: https://golang.org/pkg/crypto/rand/#Read

trace/otel_span.go
paulosman: Good catch. You can't have `{}` in a conditional, which is what threw me off originally, but there is an accessor which I can use. Updated.

trace/trace.go
paulosman: Fixed, thank you!

trace/trace.go
paulosman: Thanks for the catch and info re: `rand.Read`. That's much better. 

I went back and forth a bit on the defaults and decided that yeah, it's nice to make a best effort in the case that we can't generate a pseudo-random ID. I've updated the function to use a counter as the fallback as you suggest.

propagation/propagation.go
paulosman: Amazon Load Balancers use `Root` and `Self`:

https://docs.aws.amazon.com/elasticloadbalancing/latest/application/load-balancer-request-tracing.html

I should include support for instrumented applications though, and parse `Parent`. It also occurs to me that additional fields are allowed, so we can put those in the `TraceContext` field.

wrappers/hnynethttp/nethttp.go
```diff
@@ -156,7 +158,23 @@ func (ht *hnyTripper) spanRoundTrip(ctx context.Context, span *trace.Span, r *ht
 	}
 	span.AddField("meta.type", "http_client")
 	span.AddField("name", "http_client")
-	r.Header.Add(propagation.TracePropagationHTTPHeader, span.SerializeHeaders())
+
+	otelSpan := trace.OTelSpanFromContext(ctx)
+	ctx = oteltrace.ContextWithSpan(ctx, otelSpan)
```
toshok, line 163: This feels out of place here- shouldn't this be in a propagator?

trace/context.go
```diff
@@ -34,6 +35,33 @@ func PutTraceInContext(ctx context.Context, trace *Trace) context.Context {
 	return context.WithValue(ctx, honeyTraceContextKey, trace)
 }
 
+// GetRemoteSpanContextFromContext retrieves a SpanContext form the passed in
```
toshok, line 38:
```suggestion
// GetRemoteSpanContextFromContext retrieves a SpanContext from the passed in
```

trace/trace.go
```diff
@@ -100,15 +138,15 @@ func (t *Trace) AddField(key string, val interface{}) {
 // The serialized form may be passed to NewTrace() in order to create a new
```
toshok, line 138: this comment doesn't line up with reality anymore, does it?

actually - what calls this function now - feels superfluous now that we have propagators?

trace/trace.go
```diff
@@ -172,6 +210,68 @@ func (t *Trace) Send() {
...
+	TraceID      string
+	ParentID     string
+	TraceContext map[string]interface{}
+	Dataset      string
+	TraceState   byte
+}
+
+// SerializeHeader returns a string representation, currently in Honeycomb trace
+// context header format.
+func (sc *SpanContext) SerializeHeader() string {
```
toshok, line 226: ideally this wouldn't be exported from this package and would move to the honeycomb propagator instead.

trace/trace.go
```diff
@@ -472,3 +580,39 @@ func (s *Span) createChildSpan(ctx context.Context, async bool) (context.Context
...
+	}
+}
+
+// build generates random bytes and encodes them as a hex string. The
+// length of the id depends on the length of the byte slice provided. If
+// rand.Read fails, the result will be a best attempt based on a counter.
+func (r *randomIdentifier) build(id []byte) string {
+	_, err := r.randomBytes(id)
+	if err != nil {
+		nonRandomId := make([]byte, len(id))
```
toshok, line 604: why can't we use continue to use `id` here for this (instead of creating a new `[]byte`)?

trace/trace.go
```diff
@@ -472,3 +580,39 @@ func (s *Span) createChildSpan(ctx context.Context, async bool) (context.Context
...
+
+type randomIdentifier struct {
+	randomBytes   func([]byte) (int, error)
+	incrementLock sync.Mutex
+	nonRandomId   uint64
+}
+
+// newRandomIdentifier returns a new randomIdentifier with the default pseudo-random
+// generator.
+func newRandomIdentifier() *randomIdentifier {
```
toshok, line 592: naming nit - the name of this function makes me think it's going to return an ID itself.  but it doesn't, it returns an ID generator.

trace/trace.go
```diff
@@ -472,3 +580,39 @@ func (s *Span) createChildSpan(ctx context.Context, async bool) (context.Context
...
+	return &randomIdentifier{
+		randomBytes: rand.Read,
```
toshok, line 594: maybe do an initial `rand.Uint64()` to initialize `nonRandomId`?  or do we really always want it to iterate starting at 0 (outside of tests)?

trace/trace.go
```diff
@@ -472,3 +580,39 @@ func (s *Span) createChildSpan(ctx context.Context, async bool) (context.Context
...
+}
+
+// build generates random bytes and encodes them as a hex string. The
+// length of the id depends on the length of the byte slice provided. If
+// rand.Read fails, the result will be a best attempt based on a counter.
+func (r *randomIdentifier) build(id []byte) string {
+	_, err := r.randomBytes(id)
+	if err != nil {
+		nonRandomId := make([]byte, len(id))
+		binary.BigEndian.PutUint64(nonRandomId, r.getIncrementedIdCounter())
```
toshok, line 605: this will panic if len(id) < 8, which we might want to guard against.  it'd be a bug within the beeline code, but it'd still potentially cause problems for user code.

wrappers/common/common.go
```diff
@@ -46,10 +46,22 @@ func StartSpanOrTraceFromHTTP(r *http.Request) (context.Context, *trace.Span) {
 	ctx := r.Context()
 	span := trace.GetSpanFromContext(ctx)
 	if span == nil {
-		// there is no trace yet. We should make one! and use the root span.
-		beelineHeader := r.Header.Get(propagation.TracePropagationHTTPHeader)
+		var spanContext *trace.SpanContext
+		if trace.GlobalConfig.TraceHTTPHeaderParserHook != nil {
```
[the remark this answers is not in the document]
paulosman: Good call. I was following a pattern here, but I don't think it buys us anything. Also, renaming `HoneycombContext` and `AmazonContext` to `HoneycombHTTPPropagator` and `AmazonHTTPPropagator` makes the code a lot clearer. Fixed.

trace/trace.go
paulosman: Great catch! Added a guard.

trace/trace.go
paulosman: Renamed to `newIDGenerator` and renamed the struct to `idGenerator`.

trace/trace.go
paulosman: Ah, I love it when there's code that can be deleted. Yup - removed.

propagation/propagation.go
```diff
@@ -1,119 +1,209 @@
...
+// PAYLOAD is a list of comma-separated params (k=v pairs), with no spaces.  recognized
+// keys + value types:
+//
+//  trace_id=${traceId}    - traceId is an opaque ascii string which shall not include ','
+//  parent_id=${spanId}    - spanId is an opaque ascii string which shall not include ','
+//  dataset=${datasetId}   - datasetId is the slug for the honeycomb dataset to which downstream spans should be sent; shall not include ','
+//  context=${contextBlob} - contextBlob is a base64 encoded json object.
+//
+// ex: X-Honeycomb-Trace: 1;trace_id=weofijwoeifj,parent_id=owefjoweifj,context=SGVsbG8gV29ybGQ=
+type HoneycombContext struct{}
```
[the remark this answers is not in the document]
paulosman: Good call. I was following a pattern here, but I don't think it buys us anything. Also, renaming `HoneycombContext` and `AmazonContext` to `HoneycombHTTPPropagator` and `AmazonHTTPPropagator` makes the code a lot clearer. Fixed.

wrappers/common/common.go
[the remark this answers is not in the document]
paulosman: Much much clearer. Thank you.

wrappers/hnynethttp/nethttp.go
paulosman: Oooh, yeah they should be. The reason they're here is because the W3C trace context propagator in the otel sdk expects an otel span, so we grab it here and shove it into the context before calling the Inject method on each of the propagators, one of which could be the W3C propagator in the otel sdk. 

This is making me think it'd be better to wrap that propagator. Unfortunately, that'd mean wrapping all of the otel propagators, which feels brittle. Open to thoughts - better to just comment here, or move these into a wrapped propagator?

trace/trace.go
paulosman: Yeah, that's much better. Thanks.

wrappers/hnynethttp/nethttp.go
paulosman: Opted to keep as-is, but with better commenting.

paulosman: With the most recent commit, this is gone!

trace/trace.go
```diff
@@ -37,40 +73,51 @@ type Trace struct {
...
+	rootSpan.spanContext = &SpanContext{
+		ParentID: rootSpan.spanID,
+		TraceID:  trace.traceID,
+	}
+
+	if sc != nil {
+		rootSpan.spanContext.Dataset = sc.Dataset
+		rootSpan.spanContext.TraceContext = sc.TraceContext
+		rootSpan.spanContext.TraceState = sc.TraceState
+	}
```
maplebed, line 119: This makes me think that there are two copies of trace context - one on the span on one in the trace. Do we need both? Spans have a link to the trace object they can use to get a trace level fields. I'll keep reading to try and understand better how the span context on a span is used

trace/trace.go
```diff
@@ -172,6 +201,42 @@ func (t *Trace) Send() {
...
 
+// SpanContext contains information related to a Span. Context should be carried
+// around by the currently active span, and optionally propagated in trace context
+// headers.
+type SpanContext struct {
+	TraceID      string
+	ParentID     string
+	TraceContext map[string]interface{}
+	Dataset      string
+	TraceState   byte
```
maplebed, line 212: Need a comment on TraceState describing what it is. Maybe the others too? But this one for sure.

trace/trace.go
```diff
@@ -190,6 +255,12 @@ type Span struct {
 	trace        *Trace
 	eventLock    sync.Mutex
 	sendLock     sync.RWMutex
+	spanContext  *SpanContext
```
maplebed, line 258: I'm leery of having two copies of the span's metadata here - what do we do if the `spanContext` trace ID and the `Span`'s trace ID conflict?

wrappers/common/common.go
```diff
@@ -42,14 +42,28 @@ func NewResponseWriter(w http.ResponseWriter) *ResponseWriter {
...
 	ctx := r.Context()
 	span := trace.GetSpanFromContext(ctx)
 	if span == nil {
-		// there is no trace yet. We should make one! and use the root span.
-		beelineHeader := r.Header.Get(propagation.TracePropagationHTTPHeader)
+		var spanContext *trace.SpanContext
+		for _, propagator := range getParsersForRequest(ctx, r) {
+			spanContext = propagator.Parse(ctx, r.Header)
+			spanContext = spanContext.MergeSpanContext(
+				trace.GetRemoteSpanContextFromContext(ctx),
```
maplebed, line 62: This behavior feels like a trap to me. I think we should accept all data from a single source and use that. If we want to examine multiple sources, that seems fine, but combining data from those sources feels like it will put us in a dangerous position. I would prefer to see us be clear about the order of importance if multiple headers are present and have a "most important wins" strategy.

Is there a user story style description of the circumstances in which merging data from multiple headers leads to the desired outcome?

beeline.go
```diff
@@ -38,21 +39,39 @@ type Config struct {
...
+	// TraceHeaderParserHook is a function call that will be invoked with the context
+	// and HTTP request on all incoming HTTP requests. The function registered must
+	// return a list of HTTPPropagator objects that will be used to extract trace
+	// context from headers attached to the incoming request. Each propagator will
+	// be used, in sequence, and any conflicting information will be squashed by
+	// the later propagator. Decisions about which propagators to return can be
+	// made based on information in the context or HTTP request. The function may
+	// also return an empty array, which will result in no incoming headers being
+	// parsed.
+	TraceHTTPHeaderParserHook func(ctx context.Context, r *http.Request) []trace.HTTPPropagator
```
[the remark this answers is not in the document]
maplebed: Should add to this comment that if left undefined, Honeycomb trace headers will be honored.

beeline.go
```diff
@@ -38,21 +39,39 @@ type Config struct {
...
+	// TraceHTTPHeaderPropagationHook is a function call that will be invoked with
+	// the context and HTTP request on all outgoing HTTP requests. The function
+	// registered must return a list of HTTPPropagator objects that will be used
+	// to inject trace context into the headers of the outgoing request. Decisions
+	// about what wire format to use and whether to inject headers at all can be
+	// made based on infomration in the context or HTTP request. The function may
+	// also return an empty array, which will result in no trace propagation headers
+	// being included in outgoing requests.
+	TraceHTTPHeaderPropagationHook func(ctx context.Context, r *http.Request) []trace.HTTPPropagator
```
maplebed, line 74: Should add to this comment that if left undefined, Honeycomb trace headers will be propagated.

beeline.go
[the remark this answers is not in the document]
maplebed: Should add that these functions are only invoked if you use one of Honeycomb's wrappers (either `hnynethttp` or one of the web framework wrappers). 

I think this bolsters the argument that this config might be best done at the level of the wrapper rather than at the `beeline` level.

beeline.go
maplebed: Similar to above, these hooks will only be used if you've  used `hnynethttp.WrapRoundTripper` to wrap your HTTP client's `RoundTripper` implementation. I think this may be a sign that the config might be better pushed closer to that package.

trace/trace.go
```diff
@@ -2,25 +2,61 @@ package trace
...
+// HeaderSupplier is an interface that specifies methods for storing and retrieving
+// key value pairs. It is implemented by http.Headers.
+type HeaderSupplier interface {
+	Set(key string, value string)
+	Get(key string) string
+}
+
+// HTTPPropagator wraps the interface with the same name from the OpenTelemetry Go SDK.
+// It intentionally uses different method names to highlight the fact that it's API is
+// different than the analagous object in the OpenTelemetry Go SDK.
```
maplebed, line 31: If it's not the same interface, it feels misleading to give it the same name. I'm not sure what name to use; some brainstorming will come up with something, I'm sure.

trace/trace.go
```diff
@@ -2,25 +2,61 @@ package trace
...
+type HTTPPropagator interface {
+	// Parse is responsible for retrieving the appropriate header(s) from header and
+	// using the information contained within to create and return a SpanContext object.
+	Parse(ctx context.Context, header HeaderSupplier) *SpanContext
+	// Insert is responsible for retrieving the appropriate information about the current
+	// span from ctx and using the information contained within to generate a serialized
+	// header(s) which are then inserted into header.
+	Insert(ctx context.Context, header HeaderSupplier)
```
maplebed, line 39: Why is it a `context.Context` here instead of a `SpanContext`? I'm trying to make `Parse` and `Insert` mirrors of each other and struggling.

beeline.go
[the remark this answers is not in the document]
paulosman: > This is the first introduction of anything specific to HTTP in to the beeline itself. It makes me wonder if these hooks should be part of the middleware wrappers that are automatic instrumentation specific to HTTP rather than in the main beeline package.
> So far the way those breakdown:
> 
> * beeline - dirt simple interface to adding tracing data to an application
> * trace - more complete and powerful and complicated interface to traces
> * autoinstrumentation - hooks for specific packages, including HTTP
> 
> In the interests of "dirt simple" it might be right to move HTTP into the beeline but I'd like to talk through it a bit more.

I'm going to write up some example code that uses some possible interface definitions - I think it'd be easier to discuss this if we had something tangible to point at.

trace/trace.go
paulosman: I think we need to tease out the responsibility of `Trace` and `Span` a little bit more here. Again, probably better with something in a doc that describes the relationship between each of these in the context of a Honeycomb event. I'll work on that.

wrappers/common/common.go
paulosman: Clearest boundary:

- Multiple incoming headers, multiple propagators defined. 
- Last writer wins. Last defined propagator that has header(s) present is what will be used.
- Document the crap out of this in the hook definition.

Use case: 

- A service sits behind an ALB and that is how it is accessed from the public internet. That service can also be called directly from peer services within the same VPC. In this case, if a Honeycomb header exists, and NO amazon header exists, we should use it. If both exist, the Amazon header should clearly win.

trace/trace.go
```diff
@@ -472,3 +535,42 @@ func (s *Span) createChildSpan(ctx context.Context, async bool) (context.Context
 	ctx = PutSpanInContext(ctx, newSpan)
 	return ctx, newSpan
 }
+
+type idGenerator struct {
+	randomBytes   func([]byte) (int, error)
+	incrementLock sync.Mutex
+	nonRandomID   uint64
+}
```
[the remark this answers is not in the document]
maplebed: I'd like to rephrase that comment - there's some kind of generalization showing up here and I'm not sure why it's here. I think the task being fulfilled is "Please give me an ID of X length". Can we make the API for ID generation closer to that? Do we need generators or can this be a function call such as `trace.GetNewID(8)` to get an 8 character ID? Is that oversimplifying it in a way that will make other things more difficult? (I wonder about whether it's particularly useful to be able to use different ID generation in tests, for example.)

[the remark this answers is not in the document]
maplebed: Ooh, and that question about testing highlights a different change that might happen here. If we do want to be able to replace ID generation for testing, we'll want to use an interface here instead of a struct. And if we don't, there's room to simplify the calling code.

paulosman:
Closing this PR as it has grown too large and should be broken up. Thank you everyone for the extremely valuable feedback. This PR will be referenced in several new PRs that attempt to break this change into incremental steps.

PR #110, opened 2020-06-19
Replace UUID strings with W3C Trace Context compatible IDs.

paulosman:
The [W3C Trace Context](https://www.w3.org/TR/trace-context/) specification defines span ids as 8-byte arrays and trace ids as 16-byte arrays. In order to maintain compatibility with other services using that standard, we should start generating our ids in the same way. This will make it easier to add trace propagation interoperability between OpenTelemetry and Beeline instrumented services.

This is part of a series of PRs meant to replace #108.

### DIFF
--- a/trace/trace.go
+++ b/trace/trace.go
@@ -45,8 +45,8 @@ type Trace struct {
 	traceLevelFields map[string]interface{}
 }
 
-// getNewID generates a hex encoded string with the specified number of bytes.
-// It is used for ID generation for traces and spans.
+// getNewID generates a lowercase hex encoded string with the specified number
+// of bytes. It is used for ID generation for traces and spans.
 func getNewID(length uint16) string {
 	id := make([]byte, length)
 	// rand.Seed is called in libhoney's init, so this is sure to have well-seeded random content.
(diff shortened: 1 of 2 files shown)


trace/trace.go
```diff
@@ -39,6 +45,14 @@ type Trace struct {
 	traceLevelFields map[string]interface{}
 }
 
+// getNewID generates a hex encoded string with the specified number of bytes.
+// It is used for ID generation for traces and spans.
+func getNewID(length uint16) string {
+	id := make([]byte, length)
+	_, _ = rand.Read(id)
```
maplebed, line 52:
```suggestion
	// rand.Seed is called in libhoney's init, so this is sure to have well-seeded random content.
	_, _ = rand.Read(id)
```

trace/trace.go
```diff
@@ -39,6 +45,14 @@ type Trace struct {
 	traceLevelFields map[string]interface{}
 }
 
+// getNewID generates a hex encoded string with the specified number of bytes.
+// It is used for ID generation for traces and spans.
```
maplebed, line 49:
```suggestion
// getNewID generates a lowercase hex encoded string with the specified number
// of bytes. It is used for ID generation for traces and spans.
```

PR #111, opened 2020-06-24
Generalize PropagationContext and Related Methods

paulosman:
In preparation for adding support for different trace propagation wire formats, this PR generalizes PropagationContext to include fields found in all of the formats we plan to support. It also renames generic functions like `MarshalTraceContext` to the more specific `MarshalHoneycombTraceContext`. (In future PRs we plan to introduce marshal / unmarshal functions for other trace header formats). The old function names are left as wrappers and are marked as deprecated.

This also adds a `NewTraceFromPropagationContext` function to the `trace` package. This function will be used to create traces from newly parsed headers but, unlike `NewTraceFromSerializedHeaders`, has no knowledge of the header format itself. `NewTrace` wraps `NewTraceFromSerializedHeaders` but may in the future change to wrap `NewTraceFromPropagationContext`.

Looking for 👀 from @maplebed. Tagging @katiebayes and @dsoo as an FYI and for additional 👀 but not required.

### DIFF
--- a/propagation/honeycomb.go
+++ b/propagation/honeycomb.go
@@ -65,7 +65,7 @@ func UnmarshalHoneycombTraceContext(header string) (*PropagationContext, error) 
 	// pull the version out of the header
 	getVer := strings.SplitN(header, ";", 2)
 	if getVer[0] == "1" {
-		return UnmarshalTraceContextV1(getVer[1])
+		return UnmarshalHoneycombTraceContextV1(getVer[1])
 	}
 	return nil, &PropagationError{fmt.Sprintf("unrecognized version for trace header %s", getVer[0]), nil}
 }
--- a/propagation/honeycomb.go
+++ b/propagation/honeycomb.go
@@ -1,0 +1,105 @@
+package propagation
+
+import (
+	"encoding/base64"
+	"encoding/json"
+	"fmt"
+	"net/url"
+	"strings"
+)
+
+// assumes a header of the form:
+
+// VERSION;PAYLOAD
+
+// VERSION=1
+// =========
+// PAYLOAD is a list of comma-separated params (k=v pairs), with no spaces.  recognized
+// keys + value types:
+//
+//  trace_id=${traceId}    - traceId is an opaque ascii string which shall not include ','
+//  parent_id=${spanId}    - spanId is an opaque ascii string which shall not include ','
+//  dataset=${datasetId}   - datasetId is the slug for the honeycomb dataset to which downstream spans should be sent; shall not include ','
+//  context=${contextBlob} - contextBlob is a base64 encoded json object.
+//
+// ex: X-Honeycomb-Trace: 1;trace_id=weofijwoeifj,parent_id=owefjoweifj,context=SGVsbG8gV29ybGQ=
+
+const (
+	TracePropagationHTTPHeader = "X-Honeycomb-Trace"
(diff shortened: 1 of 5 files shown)


propagation/honeycomb.go
```diff
@@ -0,0 +1,105 @@
...
+// ex: X-Honeycomb-Trace: 1;trace_id=weofijwoeifj,parent_id=owefjoweifj,context=SGVsbG8gV29ybGQ=
+
+const (
+	TracePropagationHTTPHeader = "X-Honeycomb-Trace"
+	TracePropagationVersion    = 1
+)
+
+// MarshalHoneycombTraceContext uses the information in prop to create a trace context header
+// in the Honeycomb trace header format. It returns the header and header name as a key value
+// pair in a map.
```
maplebed, line 34:
```suggestion
// in the Honeycomb trace header format. It returns the serialized form of the trace context, 
// ready to be inserted into the headers of an outbound HTTP request.
```

propagation/honeycomb.go
```diff
@@ -0,0 +1,105 @@
...
+func MarshalHoneycombTraceContext(prop *PropagationContext) string {
```
maplebed, line 35: `prop` is a pointer to a `PropagationContext` so this code will panic when it gets to `prop.TraceContext` if it gets passed `nil`. 
(note that this was a pre-existing bug, so I'm glad that nobody's called this function with a nil PropagationContext before!)
```suggestion
func MarshalHoneycombTraceContext(prop *PropagationContext) string {
	if prop == nil {
		return ""
	}
```

propagation/honeycomb.go
```diff
@@ -0,0 +1,105 @@
...
+	)
+}
+
+// UnmarshalHoneycombTraceContext parses the information provided in header and creates a
+// PropagationContext instance.
+func UnmarshalHoneycombTraceContext(header string) (*PropagationContext, error) {
+	// pull the version out of the header
+	getVer := strings.SplitN(header, ";", 2)
+	if getVer[0] == "1" {
+		return UnmarshalTraceContextV1(getVer[1])
```
maplebed, line 65:
```suggestion
		return UnmarshalHoneycombTraceContextV1(getVer[1])
```

propagation/propagation.go
```diff
@@ -1,119 +1,59 @@
...
-	"encoding/base64"
-	"encoding/json"
 	"fmt"
-	"net/url"
-	"strings"
 )
 
-// assumes a header of the form:
+// Default to using the Honeycomb Trace Context header.
+const defaultTracePropagationHTTPHeader = TracePropagationHTTPHeader
```
maplebed, line 8: I'm not sure what this line is for. Perhaps it's intended for a future PR?

paulosman: That's correct - I'm going to remove it from this PR though - meant to remove it before pushing.

paulosman:
> I appreciate the `type Propagation = PropagationContext` line in there. I'm not sure it will do enough to accommodate the function signatures in here changing from `*Propagation` to `*PropagationContext`. That said, I'm not sure I care. I'm willing to do the careful migration path for changing arguments that are really different (a serialized string vs. a struct). I don't think it's as necessary when changing the name of a struct that retains the same contents.
> 
> LGTM.

I was dubious too, but it seems to work:

https://play.golang.org/p/mOwzhJUJZAD

Can't hurt!

PR #112, opened 2020-06-25
Add Middleware for gin-gonic

Nalum:
I'm adding this package to support https://github.com/gin-gonic/gin

I'm not sure if the approach taken is the best or not.

At the moment I'm just looking for feedback on it and some help in regards to getting it into a place that is usable generally by the wider community.

I based this initial work on the httprouter package as gin itself uses that.


wrappers/hnygingonic/gingonic.go
```diff
@@ -0,0 +1,76 @@
...
+		// pull out any GET query params
+		for key, value := range c.Request.URL.Query() {
+			if len(value) > 1 {
+				span.AddField("handler.query."+key, value)
+			} else if len(value) == 1 {
+				span.AddField("handler.query."+key, value[0])
+			} else {
+				span.AddField("handler.query."+key, nil)
+			}
+		}
```
maplebed, line 42: This is tempting and dangerous for any publicly accessible web services. All the spew on the internet that attempts to hit your webserver with paths like ../../../../../../password.txt or whatevs also often throws garbage query parameters in your URLs. By automatically accepting all query parameters as fields you are ceding control of your dataset schema to the public internet - a recipe for disaster. I say this as one who has fallen for this trap at least 3 times. A far more stable approach is to explicitly add only the parameters that you care about when you start to process them and accept that it does require a little more work throughout the rest of your codebase.

There's one exception to this advice, and I don't know whether gin does this - but it's the reason this is down on this block intsead of above on the `c.Params` block. If the contents of `c.Params` does not come from the list of query parameters in the URL but instead is registered with gin as part of the handler creation, then this is totally fine since they've effectively already been explicitly declared as desired arguments. However, I'm pretty sure this block is grabbing things from the URL itself, so is probably not safe.

Nalum: Ah, that totally makes sense, I'll update that. My initial thoughts on it would be to add a param to the `Middleware` function that took the GET params you expect to deal with in your requests.

e.g.

```go
func Middleware(getParams map[string]struct{}) gin.HandlerFunc {
	return func(c *gin.Context) {
		...
		// pull out any GET query params
		if getParams != nil {
			for key, value := range c.Request.URL.Query() {
				if _, ok := getParams[key]; ok {
					if len(value) > 1 {
						span.AddField("handler.query."+key, value)
					} else if len(value) == 1 {
						span.AddField("handler.query."+key, value[0])
					} else {
						span.AddField("handler.query."+key, nil)
					}
				}
			}
		}
		...
	}
}
```

As for the `c.Params` I believe that is setup from the url string defined in the router

e.g.

```go
func main() {
	...
	router.GET("/person/:name/pets", listPets) // In this route :name is a param
	...
}
```

wrappers/hnygingonic/hnygingonic_example_router_test.go
```diff
@@ -0,0 +1,79 @@
...
+	hnyctx, span := StartSpan(c, "main.home")
+	defer span.Send()
+	span.AddField("Welcome", "Home")
+	childFunction(hnyctx)
+	c.Data(http.StatusOK, "text/plain", []byte(`Welcome Home`))
+}
+
+func alive(c *gin.Context) {
+	_, span := StartSpan(c, "main.alive")
+	defer span.Send()
```
maplebed, line 52: I think I would leave out these two lines here. The middleware has already created a root span - these lines will add a child span. Which is fine, when it's useful, and it's nice to have the example that you can still create your own spans (and how) in the "complicated" handlers like `hme` or `ready`, but it's also fine to leave them out. For the health check leaving it out would be a good way of saying "yeah, we really don't need it here."

Nalum: Thanks :+1: I have updated the example.

wrappers/hnygingonic/gingonic.go
```diff
@@ -0,0 +1,82 @@
...
+		// push the context with our trace and span on to the request
+		c.Request = c.Request.WithContext(ctx)
+
+		// replace the writer with our wrapper to catch the status code
+		wrappedWriter := common.NewResponseWriter(c.Writer)
+
+		// pull out any variables in the URL, add the thing we're matching, etc.
+		for _, param := range c.Params {
+			span.AddField("handler.vars."+param.Key, param.Value)
+		}
```
Nalum, line 30: Does it make sense for this (and below in `Middleware`) to be `AddField` or would it be better for it to be `AddTraceField`?

Nalum:
@maplebed @paulosman just occured to me that it might be worth doing what is done with GET params with Headers as well, what do you think?

wrappers/hnygingonic/gingonic.go
maplebed: I can see the lure. My instinct is to leave trace level fields for the application developers to choose what gets threaded through all spans rather than having middleware do it (easy enough to add the same field within the handler) but I don't have compelling experience pushing me one way or the other.

maplebed:
> @maplebed @paulosman just occured to me that it might be worth doing what is done with GET params with Headers as well, what do you think?

This is an interesting idea. I wouldn't put it in this PR, but with the work currently ongoing to increase trace header interop between the beelines and other tracing libraries (eg #113 and its coming followers) there is now room for additional configuration on a per-middleware basis. We do frequently get requests to add specific headers (eg x-forwarded-proto in addition to x-forwarded-for) to various middleware packages and using the same configuration style for trace header interop and other configurable manipulations of the middleware would be a neat road to explore.

wrappers/hnygingonic/gingonic.go
Nalum: Understood, it also occurs to me that it could be seen as unexpected behaviour if the rest of the middleware packages are not doing it :+1:

Nalum:
@paulosman Thanks I've fixed those conflicts :+1:

PR #113, opened 2020-06-26
Add Support for Amazon and W3C Trace Context Headers

paulosman:
This change adds support for Amazon and W3C Trace Context Headers.
The following methods have been added to the propagation package:

- MarshalAmazonTraceContext(*PropagationContext) string
- UnmarshalAmazonTraceContext(string) (*PropagationContext, error)
- MarshalW3CTraceContext(context.Context, 
      *PropagationContext) (context.Context, map[string]string])
- UnmarshalW3CTraceContext(context.Context, map[string]string) 
      (context.Context, *PropagationContext, error)

These functions allow users to parse and propagate headers in Amazon or
W3C Trace Context formats.


propagation/w3c.go
```diff
@@ -0,0 +1,165 @@
...
+	"go.opentelemetry.io/otel/api/kv"
+	"go.opentelemetry.io/otel/api/trace"
+	"google.golang.org/grpc/codes"
+)
+
+// MarshalHoneycombTraceContext uses the information in prop to create trace context headers
+// that conform to the W3C Trace Context specification. The header values are set in headers,
+// which is an HTTPSupplier, an interface to which http.Header is an implementation. The headers
+// are also returned as a map[string]string.
+func MarshalW3CTraceContext(ctx context.Context, prop *PropagationContext) (context.Context, map[string]string) {
```
maplebed, line 16: Can you say something about why the function signature for marshaling w3c is different from marshaling honeycomb or aws headers? It feels like if we have several functions `MarshalFooTraceContext` they should all have the same signature to make clear that they are doing the same thing. If they are not doing the same thing, is there an abstraction layer here that's leaking? 
This may be resolved with some example code once we get further in the process...

propagation/w3c.go
```diff
@@ -0,0 +1,165 @@
...
+	return otelSpan{
+		ctx: spanCtx,
+	}, nil
+}
+
+// otelSpan is an implementation of the trace.Span interface. That interface is fairly
+// wide, so there are a lot of methods on this type that are noops. The only field we
+// need in order to use the opentelemetry-go sdk w3c trace context propagator is the
+// trace.SpanContext, so we populate that.
+type otelSpan struct {
```
maplebed, line 81:
```suggestion
// otelSpan is an implementation of the open telemetry trace.Span interface. We
// want an implementation in order to use the existing otel code for w3c trace
// propagation parsing rather than duplicating that code here. That interface is
// fairly wide, so there are a lot of methods on this type that are noops. The
// only field we need in order to use their trace context propagator is the
// trace.SpanContext, so we populate that.
type otelSpan struct {
```

propagation/propagation_test.go
```diff
@@ -55,6 +56,55 @@ func TestMarshalTraceContext(t *testing.T) {
 	assert.NoError(t, err, "roundtrip error")
 }
 
+func TestMarshalAmazonTraceContext(t *testing.T) {
+	// NOTE: we only support strings for trace context in amazon headers
```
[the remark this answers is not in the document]
paulosman: Elaborated! Hopefully it's clearer now.

propagation/propagation_test.go
```diff
@@ -146,3 +196,85 @@ func TestUnmarshalTraceContext(t *testing.T) {
 		}
 	}
 }
+
+func TestUnmarshalAmazonTraceContext(t *testing.T) {
+	testCases := []struct {
+		name       string
```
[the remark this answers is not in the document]
paulosman: ✨

propagation/w3c.go
paulosman: They're different because W3C Trace Context can produce multiple headers, whereas both the Amazon and Honeycomb marshal functions produce only one. A possible solution is to make the Amazon and Honeycomb marshal functions also return a `map[string]string`. Additionally, the W3C Trace Context functions take and return a Context because those are used by the OTel SDK. Again, could just add the context to the Amazon and Honeycomb marshal / unmarshal functions even if it's ignored.

propagation/amazon.go
```diff
@@ -0,0 +1,83 @@
...
+	// used as both the parent id and the trace id.
+	if prop.TraceID != "" && prop.ParentID == "" {
+		prop.ParentID = prop.TraceID
+	}
+
+	if prop.TraceID == "" && prop.ParentID != "" {
+		return nil, &PropagationError{"parent_id without trace_id", nil}
+	}
+
+	return prop, nil
```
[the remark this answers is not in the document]
paulosman: I opted to keep the behavior "return an empty propagationcontext object if the amazon header is missing or nil" but document it. I think there's a valid argument that "no data" is different than "invalid data". The honeycomb format specifies that there must be a version number, so no data is invalid data. If the Amazon header had the same requirement, I think it'd make sense to return an error. I don't feel strongly about it, but it made sense to me that way.

propagation/amazon.go
```diff
@@ -0,0 +1,83 @@
...
+	prop.TraceContext = make(map[string]interface{})
+	for _, segment := range segments {
+		keyval := strings.SplitN(segment, "=", 2)
+		switch strings.ToLower(keyval[0]) {
+		case "self":
+			prop.ParentID = keyval[1]
+		case "root":
+			prop.TraceID = keyval[1]
+		case "parent":
+			prop.ParentID = keyval[1]
```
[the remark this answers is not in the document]
paulosman: Thanks for catching this! I think you're right, the best solution is to just not parse the ParentID. All "extra" fields will be propagated via the TraceContext map anyway. I've updated the PR to go with that option. PTAL!

propagation/w3c.go
paulosman: Done. Also wrote up some alternative options in an internal quip document. Happy to discuss.

maplebed: On second thought, maybe it's a _good thing_ that the function signature for one is different. These are all expected to be called from within the hook (that will have access to the full request object, including its context). That the specifics of each protocol might need different things from the request seems reasonable - and provides a good example for someone writing some other serialization code later that does something else funky.
When I was asking about the abstraction layer above - the marshal / unmarshal functions are below the consistent portion of the abstraction (which is the callback hook). No leaking here! (maybe.)

propagation/propagation_test.go
```diff
@@ -57,7 +57,12 @@ func TestMarshalTraceContext(t *testing.T) {
 }
 
 func TestMarshalAmazonTraceContext(t *testing.T) {
-	// NOTE: we only support strings for trace context in amazon headers
+	// According to the documentation for load balancer request tracing:
+	// https://docs.aws.amazon.com/elasticloadbalancing/latest/application/load-balancer-request-tracing.html
+	// An application can add arbitrary fields for its own purposes. The load balancer preserves these fields
+	// but does not use them. In our implementation, we stick these fields in the TraceContext. Because of the
+	// implementation, the TraceContext only supports strings whereas in the Honeycomb header format, these
+	// fields are stored as base64 encoded JSON and therefore can support basic types like strings, booleans, etc.
```
maplebed, line 65: Ahh, thank you for this. In other places we might try and coerce number-looking things to a numeric type. eg if the header contained "foo=bar,bar=32,baz=true" we could try and cast the `32` to a float and the `true` to a bool. Either way I think we would break some people's expectations. Leaving it as a string seems like a good first choice and maybe change the behavior later. Maybe worth a mention in the description of the exported Unmarshal function?

propagation/w3c.go
```diff
@@ -32,6 +38,14 @@ func MarshalW3CTraceContext(ctx context.Context, prop *PropagationContext) (cont
...
 // which is an interface that defines Get and Set methods, http.Header is an implementation.
+//
+// Context is passed into this function and returned so that we can maintain the value of the
+// tracestate header. This is required in order to use the Propagator interface exported by the
+// OpenTelemetry Go SDK and avoid writing our own W3C Trace Context parser and serializer.
+//
+// If the headers are missing or empty strings, the propagation object will have zero values.
+// In the case of Span IDs and Trace IDs, they will be populated with 16 and 32 character strings
+// containing all zeroes.
 func UnmarshalW3CTraceContext(ctx context.Context, headers map[string]string) (context.Context, *PropagationContext) {
```
maplebed, line 49: the otel `trace` package does have an [`IsValid()`](https://pkg.go.dev/go.opentelemetry.io/otel/api/trace?tab=doc#SpanContext.IsValid) function that hangs off a `spanContext`. Do you want this unmarshal function to match the others in "if the trace context is invalid (eg it doesn't have a trace ID but does have other stuff) it will return an error"? Equally reasonable to put that responsibility on the caller.

propagation/propagation.go
```diff
@@ -1,3 +1,7 @@
+// Package propagation includes types and functions for marshalling and unmarshalling trace
+// context headers between various supported formats and an internal representation. It
+// provides support for traces that cross process boundaries with support for interoperability
+// between various kinds of trace context header formats.
 package propagation
```
maplebed, line 5: Do you think it would be worth while to add an `IsValid()` and `IsEmpty()` function to the `PropagationContext` type? Looking at the various possible return values from the unmarshaling functions, it might be kind to provide easy and obvious ways of doing extra validation on the context. I am especially eager to consider this since an "empty" context returned by the w3c unmarshal code will have `00000000000000000000000000000000` as the `TraceID`, which will prohibit you from using an equality comparison with an empty struct to test for the zero value.

propagation/w3c.go
```diff
@@ -0,0 +1,181 @@
...
+// If the headers are missing or empty strings, the propagation object will have zero values.
+// In the case of Span IDs and Trace IDs, they will be populated with 16 and 32 character strings
+// containing all zeroes.
+func UnmarshalW3CTraceContext(ctx context.Context, headers map[string]string) (context.Context, *PropagationContext) {
+	supplier := newSupplier()
+	for k, v := range headers {
+		supplier.Set(k, v)
+	}
+	propagator := trace.DefaultHTTPPropagator()
+	ctx = propagator.Extract(ctx, supplier)
```
maplebed, line 55: Since this is all within the same package, you have access to the private members of the supplier struct. This block can be changed to avoid allocating a second map and iterating the passed-in map - instead setting the supplier's `values` variable to use the passed in map directly. If you do it that way you can drop the constructor too.

Note that I'm also changing the name here so as to not shadow the type, so the diff is extended to cover where supplier is used.
```suggestion
	supp := supplier{
		values: headers,
	}
	propagator := trace.DefaultHTTPPropagator()
	ctx = propagator.Extract(ctx, supp)
```

propagation/w3c.go
```diff
@@ -0,0 +1,181 @@
...
+	values map[string]string
+}
+
+// newSupplier creates and returns an empty supplier with an initialized map for values.
+func newSupplier() *supplier {
+	m := &supplier{}
+	m.values = make(map[string]string)
+	return m
+}
+
```
maplebed, line 169: If you take the suggestion above about skipping iteration on the passed-in headers, I think this code will be unused and can be removed.
```suggestion
```

propagation/amazon.go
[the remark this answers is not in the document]
maplebed: I'm a little concerned at the variety of returned states available and the increased validation load this will put on the callers of these functions:
* when unmarshaling a honeycomb header, testing for `nil` error is enough; you can trust the prop
* when unmarshaling an amazon header, you must test for `nil` error and verify the prop isn't empty
* when unmarshaling a w3c header, there's no error but you have to test against `00000000` to see if it's empty

Each one is clearly documented, so maybe that's OK. Also, just as with function signatures, it emphasizes the reality that different serialization formats will provide different levels of insight into their success. 

Perhaps the easier answer than trying to assert consistency across the various unmarshaling implementations is to make it easier to test a `PropagationContext` to see if it's valid. Added that as a suggestion in the main propagation.go file.

propagation/honeycomb.go
```diff
@@ -71,8 +76,9 @@ func UnmarshalHoneycombTraceContext(header string) (*PropagationContext, error)
 }
 
 // UnmarshalHoneycombTraceContextV1 takes the trace header, stripped of the
-// version string, and returns the component parts. Trace ID and Parent ID
-// are both required. If either is absent a nil trace header will be returned.
+// version string, and returns the component parts. If the header includes a
+// parent id but not a trace id, or if the header contains an unparseable
+// string in the trace context, an error will be returned.
 func UnmarshalHoneycombTraceContextV1(header string) (*PropagationContext, error) {
```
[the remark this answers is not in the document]
paulosman: I cannot, and I also think it should be unexported. Not in this release though, since that would be a breaking change. Lemme mark it as deprecated though!

propagation/w3c.go
paulosman: This is the conclusion I've come to. I think the hooks need to be consistent, the marshal / unmarshal functions do not. They should be close in behavior but don't need to be consistent in the interface.

propagation/propagation.go
paulosman: I like `IsValid()`. I'm not sure how `IsEmpty()` would differ. For now, I've added `IsValid()` which just checks for the existence of a parent id and a trace id that are not the zero value.

propagation/amazon.go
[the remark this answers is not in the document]
paulosman: I'm letting this sit for a bit while I think about it. I think it's desirable to have consistent rules, and I'd ideally aim for "pass in a header that has a valid trace id and parent id or you'll get an error". Unfortunately, we've backed ourselves into a bit of a corner by allowing this with the Honeycomb header: 

https://github.com/honeycombio/beeline-go/blob/master/propagation/propagation_test.go#L51

So breaking that would be a breaking change. Not an API change, however, so we could probably get away with a major version bump and a note in the changelog about the change in behavior. I do note that we don't currently return an error in the W3C unmarshal function, so this would require changing the signature to:

`func UnmarshalW3CTraceContext(ctx context.Context, headers map[string]string) (context.Context, *PropagationContext, error)`

Which I'm not against. Just debating whether it's worth deprecating the old behavior.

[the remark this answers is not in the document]
paulosman: Okay, went with having a consistently applied restriction that you cannot create an invalid `PropagationContext` with an unmarshal function. Each of the `Unmarshal` functions now calls IsValid() and returns an error if the resulting `PropagationContext` contains neither a valid `TraceID` nor `ParentID`.

PR #117, opened 2020-07-16
[RFC] Add support for httptrace

NZKoz:
Go has built in support for some events which enable us to capture additional request information for outbound requests.  Specifically of interest to me was the DNS and pool information, but I added a few others that I could roughly make sense with.

I'm not even sure that this works, but figured I should share early and often, or potentially horrify you enough to in house this feature 😄

### DIFF
--- a/wrappers/hnynethttp/httptrace.go
+++ b/wrappers/hnynethttp/httptrace.go
@@ -1,0 +1,200 @@
+package hnynethttp
+
+import (
+	"context"
+	"crypto/tls"
+	"net/http/httptrace"
+	"strings"
+	"sync"
+
+	"github.com/honeycombio/beeline-go/trace"
+)
+
+func NewHttpClientTrace(ctx context.Context) *httptrace.ClientTrace {
+	if span := trace.GetSpanFromContext(ctx); span == nil {
+		// there's no span so just short circuit all this
+		return nil
+	}
+	tracer := newTracer(ctx)
+
+	return &httptrace.ClientTrace{
+		GetConn: func(hostPort string) {
+			tracer.GetConn(hostPort)
+		},
+		GotConn: func(info httptrace.GotConnInfo) {
+			tracer.GotConn(info)
+		},
+		DNSStart: func(info httptrace.DNSStartInfo) {
+			tracer.DNSStart(info)
+		},
+		DNSDone: func(info httptrace.DNSDoneInfo) {
+			tracer.DNSDone(info)
+		},
+		ConnectStart: func(network, addr string) {
+			tracer.ConnectStart(network, addr)
+		},
+		ConnectDone: func(network, addr string, err error) {
+			tracer.ConnectDone(network, addr, err)
+		},
+		TLSHandshakeStart: func() {
+			tracer.TLSHandshakeStart()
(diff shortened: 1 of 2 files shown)


BRMatt:
Speaking as another honeycomb user - I really like the idea of adding info from `httptrace` to http client spans, but some of our traces have hundreds or thousands of HTTP calls. If the default round tripper adds an extra 5 spans to each http client span that will have a dramatic impact on our bill 😬

lizthegrey:
Please use opentelemetry-go if you want httptrace, as it's already implemented and supported there.

maplebed:
I would love to see some of the extra juicy detail from `httptrace` show up in outbound HTTP spans! I have wanted this in the beeline since tracing was introduced. Thank you for this PR. I think I would suggest a new `RoundTripper` wrapper (maybe `RoundTripWithTrace`?) to make it optional to enable the http tracing behavior.

I agree with @BRMatt that there are times when it would be frustrating to explode span count, and I also see the value of visualizing the time spent in each phase of the HTTP connection as part of the waterfall. I wonder if it would be possible to make it a configurable flag, whether you push all fields (and create durations for each) up to the round tripper's span or make new spans for each?

lizthegrey:
Sorry for being a bit curt earlier -- to be clear, we'd welcome contributions of this feature to beeline-go _but_ there are some rough edges/edge cases with httptrace to be wary of. For instance, please read:

https://github.com/open-telemetry/opentelemetry-go/pull/113
https://github.com/open-telemetry/opentelemetry-go/issues/266
https://github.com/open-telemetry/opentelemetry-go/issues/906

wrappers/hnynethttp/httptrace.go
```diff
@@ -0,0 +1,200 @@
...
+	h.l.Lock()
+	defer h.l.Unlock()
+	if h.rootSpan == nil {
+		return
+	}
+	c, span := h.rootSpan.CreateChild(h.rootCtx)
+	span.AddField("name", "connect")
+	span.AddField("network", network)
+	span.AddField("addr", addr)
+	h.connectionCtx = c
```
lizthegrey, line 161: `Connect()` can be called more than once because Go issues parallel ipv4/ipv6 fetches. Thus, there is not _one_ `connectionCtx`, you'll be creating (and orphaning) one span for the connect that happens second.

maplebed:
The more I've sat with this idea the more it seems like a possibility to make the httptrace bits a separate `RoundTripper` wrapper. I like it for two reasons
* use is still pretty simple - nest the two round tripper wrappers
* providing configuration becomes easier - and more directly focused on _just_ the httptrace part

I don't know that it's actually feasible to do the nested wrapping - but it looks like it should?
`HTTPTraceRoundTripper(config, WrapRoundTripper(http.DefaultRoundTripper))`
something like that? I don't know which order they should go... Any thoughts on that style? Making it more like incoming HTTP middleware rather than an all-in-one thing?

MikeGoldsmith:
We’re going to close this for now. [OpenTelemetry-go](https://github.com/open-telemetry/opentelemetry-go) has support for HTTPTrace and would recommend you to use that.

PR #118, opened 2020-07-16
Create trace parser and propagation hooks.

paulosman:
Allow users of the beeline to create custom hooks that define how
request headers can be mapped to propagation context on incoming HTTP
requests and how a propagation context can map to trace context headers
on outgoing HTTP requests.

The parser hook is used for incoming requests and is therefore passed to
the `WrapHandlerWithConfig` function in the `hnynethttp` package:

```
import (
    ....
    "github.com/honeycombio/beeline-go/wrappers/config"
)

// Parse incoming AWS headers (e.g. from a load balancer)
func traceParserHook(r *http.Request) *propagation.PropagationContext {
    header := r.Header.Get("X-Amzn-Trace-Id")
    prop, err := propagation.UnmarshalAmazonTraceContext(header)
    if err != nil {
        ...
    }
    return prop
}

log.Fatal(http.ListenAndServe(":8000", hnynethttp.WrapHandlerWithConfig(mux, config.HTTPIncomingConfig{
    HTTPParserHook: traceParserHook,
}))
```

The propagation hook is used on outgoing requests and is therefore passed to
the `WrapRoundTripperWithConfig` function in the `hnynethttp` package:

```
// Send W3C Trace Context headers on outgoing responses
func propagateTraceHeaders(r *http.Request, *propagation.PropagationContext) map[string]string {
    ctx := r.Context()
    ctx, headers := propagation.MarshalW3CTraceContext(ctx, prop)
    return headers
}

client := &http.Client{
    Transport: hnynethttp.WrapRoundTripperWithConfig(http.DefaultTransport, config.HTTPOutgoingConfig({
        HTTPPropagationHook: propagateTraceHeaders,
    }))
}
```

### DIFF
--- a/trace/trace.go
+++ b/trace/trace.go
@@ -504,3 +504,14 @@ func (s *Span) createChildSpan(ctx context.Context, async bool) (context.Context
 	ctx = PutSpanInContext(ctx, newSpan)
 	return ctx, newSpan
 }
+
+// PropagationContext creates and returns a new propagation.PropagationContext using the
+// information in the current span.
+func (s *Span) PropagationContext() *propagation.PropagationContext {
+	return &propagation.PropagationContext{
+		TraceID:	  s.trace.traceID,
+		ParentID:	  s.spanID,
+		Dataset:	  s.trace.builder.Dataset,
+		TraceContext: s.trace.traceLevelFields,
+	}
+}
--- a/trace/trace.go
+++ b/trace/trace.go
@@ -509,9 +509,9 @@ func (s *Span) createChildSpan(ctx context.Context, async bool) (context.Context
 // information in the current span.
 func (s *Span) PropagationContext() *propagation.PropagationContext {
 	return &propagation.PropagationContext{
-		TraceID:	  s.trace.traceID,
-		ParentID:	  s.spanID,
-		Dataset:	  s.trace.builder.Dataset,
+		TraceID:      s.trace.traceID,
+		ParentID:     s.spanID,
+		Dataset:      s.trace.builder.Dataset,
 		TraceContext: s.trace.traceLevelFields,
 	}
 }
--- a/wrappers/config/config.go
+++ b/wrappers/config/config.go
@@ -1,0 +1,33 @@
+package config
+
+import (
+	"net/http"
+	"github.com/honeycombio/beeline-go/propagation"
+)
+
+// HTTPTraceParserHook is a function that will be invoked on all incoming HTTP requests
+// when it is passed as a parameter to an http.Handler wrapper function such as the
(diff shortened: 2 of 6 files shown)


wrappers/hnynethttp/nethttp.go
```diff
@@ -156,7 +173,16 @@ func (ht *hnyTripper) spanRoundTrip(ctx context.Context, span *trace.Span, r *ht
 	}
 	span.AddField("meta.type", "http_client")
 	span.AddField("name", "http_client")
-	r.Header.Add(propagation.TracePropagationHTTPHeader, span.SerializeHeaders())
+	if ht.propagationHook == nil {
```
maplebed, line 176: comment here would be great
```suggestion
	if ht.propagationHook == nil {
		// if no propagation is set, default to the Honeycomb header format
```

trace/trace.go
```diff
@@ -504,3 +504,14 @@ func (s *Span) createChildSpan(ctx context.Context, async bool) (context.Context
...
+// PropagationContext creates and returns a new propagation.PropagationContext using the
+// information in the current span.
+func (s *Span) PropagationContext() *propagation.PropagationContext {
+	return &propagation.PropagationContext{
+		TraceID:      s.trace.traceID,
+		ParentID:     s.spanID,
+		Dataset:      s.trace.builder.Dataset,
+		TraceContext: s.trace.traceLevelFields,
+	}
+}
```
maplebed, line 517: went from "why is this here?" to "right, because _everybody wants access to the propagation context_ and is seriously annoyed when the only format available is the serialized format. 👍

wrappers/config/config.go
```diff
@@ -0,0 +1,33 @@
...
+// to header values that will be injected in the outgoing request. The information in
+// the provided http.Request can be used to make decisions about what headers to include
+// in the outgoing request, for example based on the hostname of the target of the request.
+// The information in the provided PropagationContext should be used to create the serialized
+// header values. It is expected that this hook will use one of the marshal functions exported
+// in the propagation package for a number of supported formats (e.g. Honeycomb, AWS,
+// W3C Trace Context, etc).
+type HTTPTracePropagationHook func(*http.Request, *propagation.PropagationContext) map[string]string
+
+// WraperConfig stores configuration options used by various wrappers.
```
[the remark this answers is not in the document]
paulosman: Thought about this a bit and I think I agree that HTTP Incoming vs HTTP Outgoing is a good level of specificity. It provides future-proofing without going overboard. I'm considering and will likely push a change with `HTTPIncomingConfig` and `HTTPOutgoingConfig` structs in the `wrappers/config` package.

PR #131, opened 2020-08-12
Protect against potential data race in working with propagation context

maplebed:
There are two ways to get propagation headers out of the go beeline. The function to serialize headers (that returns the beeline header format) and the newer request for a `PropagationContext` object which can then be serialized into custom formats using the wrapper's propagation hooks. 

Unfortunately the latter form does not protect the trace level fields map from concurrent access. The effect is that if one gets a propagation context object and simultaneously adds a trace level field, a data race on that map access can occur. 

The first commit in this PR adds a test that confirms this race can happen. Switching your git checkout to f22a86c and running `go test ./... -race` will spit out the dreaded `WARNING: DATA RACE` message. Moving back to past the second commit in this PR and re-running the tests will show that moving the portion that works on the trace level fields into the `Trace`-attached method and making a copy of the fields protects against that race. 

The interface for calling either of these methods is unchanged.

### DIFF
--- a/trace/trace.go
+++ b/trace/trace.go
@@ -132,15 +132,25 @@ func (t *Trace) AddField(key string, val interface{}) {
 // The serialized form may be passed to NewTrace() in order to create a new
 // trace that will be connected to this trace.
 func (t *Trace) serializeHeaders(spanID string) string {
-	var prop = &propagation.PropagationContext{
-		TraceID:      t.traceID,
-		ParentID:     spanID,
-		Dataset:      t.builder.Dataset,
-		TraceContext: t.traceLevelFields,
-	}
-	t.tlfLock.RLock()
-	defer t.tlfLock.RUnlock()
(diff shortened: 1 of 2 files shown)


trace/trace.go
```diff
@@ -132,15 +132,29 @@ func (t *Trace) AddField(key string, val interface{}) {
...
-	var prop = &propagation.PropagationContext{
+	prop := t.propagationContext()
+	prop.ParentID = spanID
+	return propagation.MarshalTraceContext(prop)
+}
+
+// propagationContext returns a partially populated propagation context. It only
+// has the fields that come from the trace level - after getting the returned
+// object the caller must still fill in the span ID in order to fully populate
+// the PopulationContext struct for use creating serialized headers.
```
paulosman, line 143:
```suggestion
// the PropagationContext struct for use creating serialized headers.
```

PR #135, opened 2020-08-27
Add log when event is rejected with invalid api key

MikeGoldsmith:
Add log for more descriptive message during `readResponses` when an unauthorised response is received because if an invalid API key.

Corresponding PR in libhoney-go: https://github.com/honeycombio/libhoney-go/pull/96


MikeGoldsmith:
Yeah, you would potentially see messages both from libhoney and the beeline. I chose to add both because the libhoney change is dependent on a logger being given to the `Honeycomb` struct and the `readResponses` func in the Beeline always writes to STDOUT when the `DEBUG` flag is set in the config.

I think it's unlikely both a logger and the debug flag be set in a non-dev environment.

beeline.go
```diff
@@ -278,6 +278,8 @@ func readResponses(responses chan transmission.Response) {
 				message += fmt.Sprintf(": %s", metadata)
 			}
 			fmt.Printf("%s\n", message)
+		} else if r.StatusCode == 401 {
```
paulosman, line 281:
```suggestion
		} else if r.StatusCode == http.StatusUnauthorized {
```

paulosman (approved):

PR #146, opened 2020-09-29
Implement B3 propagator

Wilhansen:
The Zipkin system uses B3 as their propagation protocol so it's kinda awkward that beeline uses zipkin but not B3 (also, ingress-nginx has yet to support W3C Trace Context).

This PR adds that propagation using B3 mainly by piggybacking on the existing W3C implementation which uses otel libs.

A sample implementation for the hooks is:

```golang
func propagateTraceHook(r *http.Request, prop *propagation.PropagationContext) map[string]string {
	ctx := r.Context()
	ctx, headers := propagation.MarshalB3TraceContext(ctx, prop)
	return headers
}

func traceHeaderParserHook(r *http.Request) *propagation.PropagationContext {
	// Capitalization of the keys is important!
	headers := map[string]string{
		"b3":                r.Header.Get("b3"),
		"x-b3-flags":        r.Header.Get("X-B3-Flags"),
		"x-b3-traceid":      r.Header.Get("X-B3-TraceId"),
		"x-b3-spanid":       r.Header.Get("X-B3-SpanId"),
		"x-b3-parentspanid": r.Header.Get("X-B3-ParentSpanId"),
		"x-b3-sampled":      r.Header.Get("X-B3-Sampled"),
	}
	ctx := r.Context()
	ctx, prop, err := propagation.UnmarshalB3TraceContext(ctx, headers)
	if err != nil && (r.Header.Get("X-B3-TraceId") != "" || r.Header.Get("b3") != "") {
		log.Printf("Warning: cannot unmarshal b3: %+v", headers)
	}
	return prop
}
```


propagation/b3.go
```diff
@@ -0,0 +1,67 @@
...
+
+import (
+	"context"
+	"fmt"
+
+	"go.opentelemetry.io/contrib/propagators/b3"
+	"go.opentelemetry.io/otel/api/trace"
+)
+
+// MarshalHoneycombTraceContext uses the information in prop to create trace context headers
```
MikeGoldsmith, line 11:
```suggestion
// MarshalB3TraceContext uses the information in prop to create trace context headers
```

MikeGoldsmith (changes requested):
Thanks for the contribution @Wilhansen - this looks like a great start!

I've added one suggestion for a comment update and I think it would be good to add some tests to verify the behaviour.

propagation/b3.go
Wilhansen: Thanks for catching this!

MikeGoldsmith (approved):
Looks great - thanks!

emorikawa:
@Wilhansen @MikeGoldsmith, could there perhaps be an issue with the propagator. See https://github.com/honeycombio/beeline-go/issues/154

Wilhansen:
The otel 0.12 libraries still use `GetAllKeys` but their version 0.13 changed it to `Fields`:

https://github.com/open-telemetry/opentelemetry-go-contrib/commit/af2dfc7baf3392e81ea4b6938b3a8a5d39cf9db7#diff-4a74b3e6075e9b4b563e1a0768042e3829b16d15916c3976138158aab3871ccd

So I can't fix one without having to upgrade otal to 0.13, I think it goes the same for the compilation error in the w3c propagator.

PR #157, opened 2020-11-02
Add DB.BindNamed based on Tx.BindNamed

matiasanaya:
This is a straight copy from `Tx.BindNamed` so that we can access `DB.BindNamed`.

### DIFF
--- a/wrappers/hnysqlx/sqlx.go
+++ b/wrappers/hnysqlx/sqlx.go
@@ -46,6 +46,22 @@ func (db *DB) GetWrappedDB() *sqlx.DB {
 	return db.wdb
 }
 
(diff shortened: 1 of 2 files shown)


MikeGoldsmith (changes requested):
Hi @matiasanaya - thanks for the contribution!

Please could you add a test to ensure this works as expected? We have some tests already setup [here](https://github.com/honeycombio/beeline-go/blob/main/wrappers/hnysqlx/sqlx_test.go).

Thanks

matiasanaya:
@MikeGoldsmith I've added a test now. Cheers!

MikeGoldsmith (approved):

MikeGoldsmith:
Thanks @matiasanaya 👍

PR #163, opened 2020-12-15
Use separate CircleCI context for public repositories

maplebed:
In order to have CirlceCI push draft releases to github there is a separate API token that allows pushing artifacts. We set up a separate CircleCI context that has an API token with that right. Trying it out to see if it works!


vreynolds (approved):
👏

PR #169, opened 2020-12-29
Adding a gRPC interceptor wrapper.

paulosman:
This allows users to instrument gRPC services by providing a wrapped interceptor when creating a server:

```
serverOpts := []grpc.ServerOption{
	grpc.UnaryInterceptor(hnygrpc.UnaryServerInterceptorWithConfig(...)),
}
server := grpc.NewServer(serverOpts...)
```

Note that at the moment, only unary RPCs are supported. Streaming gRPC clients and servers can be added later. Outgoing instrumentation is also not supported yet, hence no `config.GRPCOutgoingConfig` or propagation hook included in this PR.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -23,6 +23,7 @@ require (
 	go.opentelemetry.io/contrib/propagators v0.15.1
 	go.opentelemetry.io/otel v0.15.0
 	goji.io/v3 v3.0.0
+	google.golang.org/grpc v1.27.0
 	google.golang.org/protobuf v1.25.0 // indirect
 	gopkg.in/yaml.v2 v2.3.0 // indirect
 )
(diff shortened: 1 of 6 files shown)


vreynolds (approved):
LGTM! Curious about the testing strategy: `startSpanOrTraceFromUnaryGRPC` has a lot of logic about span/trace handling, are tests for it not worth the effort of set up?

paulosman:
> LGTM! Curious about the testing strategy: `startSpanOrTraceFromUnaryGRPC` has a lot of logic about span/trace handling, are tests for it not worth the effort of set up?

Good question! We can probably just set up some contexts with trace info - lemme try and if it gets unwieldy I'll give up :)

paulosman:
> > LGTM! Curious about the testing strategy: `startSpanOrTraceFromUnaryGRPC` has a lot of logic about span/trace handling, are tests for it not worth the effort of set up?
> 
> Good question! We can probably just set up some contexts with trace info - lemme try and if it gets unwieldy I'll give up :)

Added tests for each scenario, and caught a case sensitivity bug (of course!) :tada:

PR #196, opened 2021-03-12
Set additional response header values in hnygorilla wrapper

### LINKED ISSUE #164: hnygorilla post-handler fields
hnynethttp, post-handler, sets:
- response.status_code
- response.content_length
- response.content_type
- response.content_encoding

of these, hnygorilla only sets status_code

Should hnygorilla set the others as well? And/or perhaps these fields should be abstracted into common.NewResponseWriter?

nathancoleman:
Fixes #164.

The hnynethttp wrapper adds fields for response headers other than the status code, and the same would be useful in the hnygorilla wrapper:
https://github.com/honeycombio/beeline-go/blob/5f09432b6f3de0e6107f87d25a134b95bd94850a/wrappers/hnynethttp/nethttp.go#L67-L75

### DIFF
--- a/wrappers/hnygorilla/gorilla.go
+++ b/wrappers/hnygorilla/gorilla.go
@@ -58,6 +58,15 @@ func Middleware(handler http.Handler) http.Handler {
 		if wrappedWriter.Status == 0 {
 			wrappedWriter.Status = 200
 		}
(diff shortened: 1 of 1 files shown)


vreynolds (approved):
Awesome, thanks!

PR #209, opened 2021-04-07
Remove deprecated wrappers in prep for major release

vreynolds:
Update propagation deprecations:
* NewTrace now takes a propagationContext instead of Headers

### DIFF
--- a/trace/trace.go
+++ b/trace/trace.go
@@ -57,6 +57,8 @@ func getNewID(length uint16) string {
 
 // NewTraceFromPropagationContext creates a brand new trace. prop is optional, and if included,
 // should be populated with data from a trace context header.
+//
+// Deprecated: use NewTrace instead.
 func NewTraceFromPropagationContext(ctx context.Context, prop *propagation.PropagationContext) (context.Context, *Trace) {
 	trace := &Trace{
 		builder:          client.NewBuilder(),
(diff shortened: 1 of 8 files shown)


trace/trace.go
```diff
@@ -94,26 +94,17 @@ func NewTraceFromPropagationContext(ctx context.Context, prop *propagation.Propa
...
 
-// NewTraceFromSerializedHeaders creates a brand new trace. serializedHeaders is optional, and if
-// included, should be the header as written by trace.SerializeHeaders(). When
-// not starting from an upstream trace, pass the empty string here.
-//
-// Deprecated: users should call NewTraceFromPropagationContext instead.
-func NewTraceFromSerializedHeaders(ctx context.Context, serializedHeaders string) (context.Context, *Trace) {
+// NewTrace creates a new trace. This wraps  NewTraceFromPropagationContext.
+// 
+// Deprecated: use NewTraceFromPropagationContext instead.
```
vreynolds, line 99: @paulosman this seemed like an oversight: NewTraceFromSerializedHeaders was a new method that was deprecated instead of NewTrace, I believe. Does that sound right?

paulosman: Right - I think the idea here was that `NewTraceFromSerializedHeaders` is a method that has the same signature as NewTrace. Introducing it allows us to encourage people to migrate from `NewTrace` to `NewTraceFromSerializedHeaders` if they want to continue using the Honeycomb header format as a string, leaving us free to change the NewTrace signature to accept a PropagationContext (and wrap `NewTraceFromPropagationContext` which can _then_ be deprecated!). So in this change, I'd suggest we:

- Change `NewTrace(ctx context.Context, serializedHeaders string) (context.Context, *Trace)` to `NewTrace(ctx context.Context, prop *PropagationContext) (context.Context, *Trace)`
- Add a Deprecated tag to `NewTraceFromPropagationContext`.

It's a bit round-about, but let's people take a stepwise approach to upgrading.

paulosman (changes requested):
Left a note about some less straightforward deprecation

paulosman (approved):
:1st_place_medal:

PR #216, opened 2021-04-13
Capture correct request.host value on http client redirects

mccutchen:
I've found that when using `hnynethttp.WrapRoundTripper`, the `request.host` property is left empty on spans generated  when the http client follows 3xx redirects (as when making a GET request to a shortened URL). This change ensures that the `request.host` property populated as expected on those spans.

This snippet from the documentation for the `URL` field on the [`http.Request` struct](https://golang.org/pkg/net/http/#Request) helps explain why `req.Host` is often empty on client requests (emphasis added):

> // For client requests, the URL's Host specifies the server to
> // connect to, while the Request's Host field **_optionally_**
> // specifies the Host header value to send in the HTTP
> // request.


vreynolds (approved):
Thank you!

PR #246, opened 2021-07-13
feat: span.AddField() support adding error

llorllale:
Small tweak to allow `span.AddField()` to support adding error types, similar to `beeline.AddField()`.

### DIFF
--- a/trace/trace.go
+++ b/trace/trace.go
@@ -264,8 +264,7 @@ func (s *Span) AddField(key string, val interface{}) {
 	s.eventLock.Lock()
 	defer s.eventLock.Unlock()
 	if s.ev != nil {
-		err, ok := val.(error)
-		if ok {
+		if err, ok := val.(error); ok {
 			s.ev.AddField(key, err.Error())
 		} else {
 			s.ev.AddField(key, val)
--- a/trace/trace.go
+++ b/trace/trace.go
@@ -4,9 +4,10 @@ import (
 	"context"
 	"crypto/rand"
 	"encoding/hex"
-	"go.opentelemetry.io/otel/trace"
 	"sync"
 	"time"
+
+	"go.opentelemetry.io/otel/trace"
 
 	"github.com/honeycombio/beeline-go/client"
 	"github.com/honeycombio/beeline-go/propagation"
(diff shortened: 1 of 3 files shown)


trace/trace.go
```diff
@@ -254,13 +255,21 @@ func newSpan() *Span {
...
+// `error` then the key is set to the error's message in the span.
 func (s *Span) AddField(key string, val interface{}) {
 	// The call to event's AddField is protected by a lock, but this is not always sufficient
 	// See send for why this lock exists
 	s.eventLock.Lock()
 	defer s.eventLock.Unlock()
 	if s.ev != nil {
-		s.ev.AddField(key, val)
+		err, ok := val.(error)
+		if ok {
```
MikeGoldsmith, line 268: This could be simplified to:
```suggestion
		if err, ok := val.(error); ok {
```

trace/trace_test.go
```diff
@@ -242,6 +245,13 @@ func TestSpan(t *testing.T) {
 
 	span.AddField("f1", "v1")
 	assert.Equal(t, "v1", span.ev.Fields()["f1"].(string), "after adding a field, field should exist on the span")
+	// add an error
+	expected := errors.New("test error")
+	span.AddField("error", expected)
+	assert.Contains(t, span.ev.Fields(), "error")
+	msg, ok := span.ev.Fields()["error"].(string)
+	require.True(t, ok)
```
MikeGoldsmith, line 253: I'd prefer to use the existing assert terminology:
```suggestion
	assert.Equal(t, true, ok)
```

trace/trace_test.go
```diff
@@ -103,6 +105,7 @@ func TestAddField(t *testing.T) {
 	_, tr := NewTrace(context.Background(), nil)
 	tr.AddField("wander", "lust")
 	assert.Equal(t, "lust", tr.traceLevelFields["wander"], "AddField on a trace should add the field to the trace level fields map")
+
```
MikeGoldsmith, line 108: Unnecessary whitespace
```suggestion
```

trace/trace_test.go
```diff
@@ -14,6 +15,7 @@ import (
 	libhoney "github.com/honeycombio/libhoney-go"
 	"github.com/honeycombio/libhoney-go/transmission"
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
```
MikeGoldsmith, line 18: Don't require this import if we use assert as below.

```suggestion
```

MikeGoldsmith (changes requested):
Thanks for the contribution @llorllale - I like the change you've provided.

I've suggested a few small tweaks, but overall look good 👍🏻

llorllale:
@MikeGoldsmith applied all suggestions. I'm not sure what's wrong with the tests - I can't reproduce locally.

robbkidd:
I banged the set. (Pushed "rerun" in CI.) ✅

MikeGoldsmith (approved):
Thanks for the updates @llorllale.

PR #249, opened 2021-07-16
fix(echo): this change ensures honeycomb captures errors correctly

wolfeidau:
This follows the pattern implemented in the opentracing echo middleware which ensures the registered error handler is invoked prior to completing the trace, otherwise echo doesn't capture the actual response code to the client.

For reference this is the opentelementry code https://github.com/open-telemetry/opentelemetry-go-contrib/blob/main/instrumentation/github.com/labstack/echo/otelecho/echo.go#L79-L92

I have included a test which demonstrates the problem.


MikeGoldsmith (approved):
Looks good - thanks for the contribution @wolfeidau 👍🏻

Issue #252, opened 2021-07-21
Remove dependency on UUIDs


paulosman:
When we introduced trace context compatibility with W3C Trace Context headers, we went from using UUIDs to 16 byte arrays. It appears that we are still using UUIDs in `db/sql` and `sqlx` instrumentation. This is potentially a bug for users combining sql instrumentation with opentelemetry instrumentation (because W3C Trace Context headers can't fit a UUID).

PR #262, opened 2021-08-19
Removing dependency on google/uuid

beanieboi:

https://github.com/honeycombio/beeline-go/issues/252

removing dependency on google/uuid


vreynolds:
Thanks for taking this on @beanieboi 🙇‍♀️ It looks like your fork is behind ours, could you update?

As far as the changes, the ID generation code would be triplicated across trace/sql/sqlx. Is there a way we can reduce that copy?

vreynolds:
Closing due to inactivity. Feel free to re-open if you're still working on it!

PR #273, opened 2021-09-21
Wrappers: add cfg to disable client spans

### LINKED ISSUE #272: Separate client-span and trace-propagation concerns
**Is your feature request related to a problem? Please describe.**

Currently, the http and grpc client wrappers perform two main functions:
1) They create new spans, to which information like request endpoints, response codes, and headers can be attached
2) They propagate the trace/span metadata across the connection to their corresponding server wrappers

These two functions are coupled; you can't get one without the other.

However, the creation and sending of client spans can (especially in a microservice architecture) nearly *double* the number of spans sent as part of a trace. The information contained in a client span -- headers, endpoint, etc -- is largely duplicated in the span sent by the server wrapper. This means that a team seeking to reduce its event volume might want to remove client spans from the equation, to get an easy 50% reduction in volume with little loss in data.

**Describe the solution you'd like**

A separation of concerns in the client wrappers. If appropriately configured, they should perform either (1), or (2), or both, as the user wants.

**Describe alternatives you've considered**

We tried using our sampler to simply discard client spans. However, because they're still _created_ and have their IDs sent across the network, the corresponding server spans all have no parent. This breaks the ability of the honeycomb UI to infer causality.

aarongable:
Add a new `DisableClientSpans` config parameter to the hnynethttp and
hnygrpc wrappers. By default (when set to the Go zero-value, or when
created using the default wrapper constructors), this has no effect.
When specifically set to `true`, this config parameter suppresses the
creation of new spans by the client wrappers. The trace propagation
metadata is still sent with the request, but the span ID is the ID of
the current span, not the ID of a newly-created span.

This allows users to decide whether or not they care about client spans
independently of whether they care about trace propagation. Since client
spans can nearly double the total number of spans sent by a microservice
architecture, this can be an important knob to tune.

Fixes #272

### DIFF
--- a/wrappers/config/config.go
+++ b/wrappers/config/config.go
@@ -39,6 +39,9 @@ type HTTPIncomingConfig struct {
 // instrumented application.
 type HTTPOutgoingConfig struct {
 	HTTPPropagationHook HTTPTracePropagationHook
+	// If DisableClientSpans is true, the client wrapper will propagate existing span
+	// metadata instead of of creating and propagating a new span.
+	DisableClientSpans bool
 }
 
 // GRPCTraceParserHook is a function that will be invoked on all incoming gRPC requests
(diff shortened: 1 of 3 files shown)


aarongable:
(Closing as #272 has been closed as wont-fix.)

PR #287, opened 2021-11-05
[hnygrpc] add a human readable version of the grpc status code to auto instrumentation

maplebed:

When looking at traffic instrumented with `hnygrpc`, I break down by `handler.method` and `response.grpc_status_code`. This is all well and good, but what I get back for the status code is 0,12, 16. 

While HTTP has been around long enough that many folks know what 200 and 404 mean, I don't have the same drilled-in understanding of what grpc codes 12 and 16 are. (turns out, `Unimplemented` and `Unauthenticated`. [reference](https://pkg.go.dev/google.golang.org/grpc@v1.40.0/codes#Code))

To aid the people using this instrumentation, this change augments `response.grpc_status_code` with `response.grpc_status_message` that contains the `.String()` version of the code, so people can see `Unimplemented` next to `12` when looking at their responses.

### DIFF
--- a/wrappers/hnygrpc/grpc.go
+++ b/wrappers/hnygrpc/grpc.go
@@ -109,7 +109,9 @@ func UnaryServerInterceptorWithConfig(cfg config.GRPCIncomingConfig) grpc.UnaryS
 		if err != nil {
 			span.AddTraceField("handler_error", err.Error())
 		}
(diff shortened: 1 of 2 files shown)


MikeGoldsmith (approved):
Great change, thanks @maplebed 👍🏻

PR #301, opened 2022-02-02
Add a RebindContext method to hnysqlx.DB wrapper

### LINKED ISSUE #59: Create a `RebindCtx(context.Context, string)` for the sql & sqlx wrappers
I'm guessing it's because the `Rebind(string)` function doesn't get a context, but right now it's creating floating spans that aren't attached to a trace.

image

https://github.com/honeycombio/beeline-go/blob/master/wrappers/hnysqlx/sqlx.go#L707

paulosman:

- The `Rebind` method in the `hnysqlx` wrapper doesn't currently take a `context.Context` parameter, which results in events being sent as single-span traces outside the existing trace context:

image

- This PR adds a `RebindContext(ctx context.Context, query string) string` method to the `DB` struct. The new method just wraps `Rebind` but uses `BuildDBSpan` which can take a context. The result is that `Rebind` spans will be included in the trace context. Hooray! 

Screen Shot 2022-02-02 at 9 51 54 AM

Fixes #59

### DIFF
--- a/wrappers/hnysqlx/sqlx.go
+++ b/wrappers/hnysqlx/sqlx.go
@@ -801,6 +801,24 @@ func (db *DB) Rebind(query string) string {
 	return str
 }
 
+// not implemented in the wrapper, but provides an API that allows us to include
+// Rebind calls in the provided trace context.
+func (db *DB) RebindContext(ctx context.Context, query string) string {
+	var err error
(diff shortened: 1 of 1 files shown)


MikeGoldsmith (approved):
Thanks @paulosman 👍🏻

PR #309, opened 2022-02-11
wip e&s

JamieDanielson:

-

-

### DIFF
--- a/beeline.go
+++ b/beeline.go
@@ -16,10 +16,11 @@ import (
 )
 
 const (
-	defaultWriteKey   = "apikey-placeholder"
-	defaultDataset    = "beeline-go"
-	defaultSampleRate = 1
-	warningColor      = "\033[1;33m%s\033[0m"
+	defaultWriteKey    = "apikey-placeholder"
+	defaultDataset     = "beeline-go"
+	defaultServiceName = "unknown_service:go"
+	defaultSampleRate  = 1
+	warningColor       = "\033[1;33m%s\033[0m"
 )
 
 // Config is the place where you configure your Honeycomb write key and dataset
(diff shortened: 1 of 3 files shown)


beeline.go
```diff
@@ -10,16 +10,18 @@ import (
...
 )
 
 const (
-	defaultWriteKey   = "apikey-placeholder"
-	defaultDataset    = "beeline-go"
-	defaultSampleRate = 1
-	warningColor      = "\033[1;33m%s\033[0m"
+	defaultWriteKey    = "apikey-placeholder"
+	defaultDataset     = "beeline-go"
+	defaultServiceName = "unknown_service:go"
```
vreynolds, line 22: Can we detect process name in the fallback? https://github.com/open-telemetry/opentelemetry-go/blob/c4ba4811a2488128fbe1d6a98c06d80157f4bef1/sdk/resource/builtin.go#L96

trace/trace.go
```diff
@@ -471,22 +471,27 @@ func (s *Span) send() {
 	s.childrenLock.Lock()
 	// classify span type
 	var spanType string
+	var spanKind string
 	switch {
 	case s.isRoot:
 		if s.parentID == "" {
 			spanType = "root"
+			spanKind = "SERVER"
```
vreynolds, line 479: Is this definitely going to correctly identify server/client/internal? Sounds like we've had [issues with this code](https://honeycomb.quip.com/eCG0AzChfSe0#RDRADAoRZaO) correctly identifying "spanType". 

OTEL defers to instrumentation to identify [server](https://github.com/open-telemetry/opentelemetry-go-contrib/blob/fe426e42f5fd35422b91d463c0624cd06f0e6330/instrumentation/github.com/gorilla/mux/otelmux/mux.go#L138)/[client](https://github.com/open-telemetry/opentelemetry-go-contrib/blob/2bb24f926b80e0d13e278cbeed85f4d10ffbccb3/instrumentation/net/http/httptrace/otelhttptrace/clienttrace.go#L212) - should we use the same pattern?

JamieDanielson: Forgot we decided to update this in a separate PR

JamieDanielson: I removed from this secstion, and started with an update for the gorilla wrapper but can drop it out in next commit.

beeline.go
```diff
@@ -96,16 +99,49 @@ type Config struct {
 	PprofTagging bool
 }
 
+func IsLegacyKey(config Config) bool {
```
robbkidd, line 102: Update to Classic?

hashtag classic

JamieDanielson: Definitely! https://github.com/honeycombio/beeline-go/pull/309/commits/ccc20ea5ffebe89784928d256d722a09c8b2b25c

MikeGoldsmith (changes requested):
Looks good - just one query regarding setting service.name when initialising config

beeline.go
```diff
@@ -162,6 +198,11 @@ func Init(config Config) {
 	// add a bunch of fields
 	if config.ServiceName != "" {
 		client.AddField("service_name", config.ServiceName)
+		client.AddField("service.name", config.ServiceName)
```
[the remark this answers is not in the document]
JamieDanielson: Hm, I don't think it's being added twice as it's `service_name` vs `service.name`, but I updated for clarity, including some code comments and a change to this block.
`service_name` is added for both classic keys (if non-null) and new keys. `service.name` is added for new keys to more closely mirror OTel.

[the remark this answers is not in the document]
JamieDanielson: After discussing with the team offline, decided we're going to add fields `service_name` and `service.name` for both classic keys and new keys. This will help in mixed environments, migrations, etc. We may deprecate `service_name` at some point in the future, but in the meantime we will send both.

MikeGoldsmith (approved):

beeline.go
```diff
@@ -96,16 +99,51 @@ type Config struct {
...
+				config.ServiceName = defaultServiceName + ":go"
+			}
+		}
+
+		// non classic key will ignore dataset, warn if configured
+		if config.Dataset != "" {
+			fmt.Println("WARN: Dataset is ignored in favor of service name. Data will be sent to service name:", config.ServiceName)
+		}
+		// but set dataset based on service name
+		config.Dataset = config.ServiceName
```
JamieDanielson, line 144: This needs to check if it's unknown service and if so, truncate to `unknown_service` before sending.

MikeGoldsmith (approved):
Looks great 👍🏻

PR #311, opened 2022-03-08
output error messages to stderr

danyelf:

E&S changes print error messages to stdout; this moves them to stderr.

I didn't change the error message for beelines configured to STDOUT; that one seems like it's meant to be part of the debug stream.

Found references from #309 to `fmt.println` and changed to `fmt.Fprintln( os.stderr )`

`go test` continues to work, but seems to print the error messages to stdout still -- but that might be `go test` quirk? (Verified -- when I actually use this is in code, it works as expected.)

### DIFF
--- a/beeline.go
+++ b/beeline.go
@@ -111,12 +111,12 @@ func Init(config Config) {
 	userAgentAddition := fmt.Sprintf("beeline/%s", version)
 
 	if config.WriteKey == "" {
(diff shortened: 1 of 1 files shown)


danyelf:
I'm open on whether the error messsages at 355 should also go into stderr

vreynolds (approved):
🎊

PR #313, opened 2022-03-28
maint: add go 1.18 to CI

vreynolds:

- go 1.18 was released earlier this month

- add 1.18 to test matrix
  - 1.18 failed `./beeline.go:177:4: fmt.Println call has possible formatting directive %s`
  - removed the warning color altogether, we're not using it for any other warnings
- update default version to 1.18
- bump go orb

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,14 +1,14 @@
 version: 2.1
 
 orbs:
-  go: circleci/go@1.7.0
+  go: circleci/go@1.7.1
 
 jobs:
   test:
     parameters:
       go-version:
         type: string
-        default: "1.14"
+        default: "1.18"
     executor:
       name: go/default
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -46,6 +46,7 @@ workflows:
                 - "1.15"
                 - "1.16"
                 - "1.17"
+                - "1.18"
```
robbkidd, line 49: Gosh, I think I like our [one-liner matrix-params pattern for versions](https://github.com/honeycombio/libhoney-go/pull/172/files#diff-78a8a19706dbd2a4425dd72bdab0502ed7a2cef16365ab7030a5a0588927bf47R23) better.

robbkidd (approved):
checks out

PR #314, opened 2022-03-30
Update several dependencies

### LINKED ISSUE #312: update pop to v6
[pop](https://github.com/gobuffalo/pop) v5 has transitive dependencies (bluemonday) that have known vulnerabilities. upgrade to v6

we'll need to drop go 1.14 from support

(from a cursory look, it seems they still have not upgraded all the things, so we may still see an alert, but at least now dependabot can keep us up to date on v6)

kentquirk:
Several dependencies had security updates; this bumps them all.

- Closes #219 
- Closes #242 
- Closes #299 
- Closes #300
- Closes #312

- #219 required updating pop from v5 to v6 (see #312), which also means that this beeline no longer works with versions of go prior to 1.16. 
- The rest of them were clean version bumps with no behavior changes.
- All unit tests still pass.


MikeGoldsmith:
Looks like 1.15 also failed tests

MikeGoldsmith (changes requested):
We should remove go 1.14 from tests to allow this to merge. Also looks like go 1.15 failed tests.

kentquirk:
It turns out that the pop upgrade to v6 depends on io/fs, which was introduced in go 1.16, so we can't support 1.15 either. I think that's OK -- it's more than a year old at this point.

JamieDanielson (commented):
> which also means that this beeline no longer works with versions of go prior to 1.16.

I added the "bump major" label to this since it is breaking

MikeGoldsmith (approved):

PR #329, opened 2022-06-03
[optimization] unconditional string construction

lizthegrey:
Follow-up to #328 as I realised that the Go compiler may need hinting about the `"app." +` happening always for it to compile time simplify.


lizthegrey:
ran some tests, it doesn't compile time inline, but it's still going to be faster.

PR #341, opened 2022-08-23
Add more request header information to GRPC handler spans

### LINKED ISSUE #339: GRPC wrapper does not appear to capture standard request headers
**Is your feature request related to a problem? Please describe.**

Spans for wrapper GRPC request handler functions don't have common request headers included.

**Describe the solution you'd like**

Specifically, `x-forwarded-for` if it is available, but consider what other headers appear commonly on GRPC requests. 

Prior art:

* our HTTP wrapper's use of [GetRequestProps](https://github.com/honeycombio/beeline-go/blob/v1.9.0/wrappers/common/common.go#L97)
* OTel's GRPC instrumentation?

robbkidd:

- Closes #339 

- Adds the following fields to a GRPC handler span:
  - request.header.x_forwarded_for
  - request.header.x_forwarded_proto
  - request.remote_addr

- Refactored the existing if-chain to a loop of headers to check and add to span if present.

### DIFF
--- a/wrappers/hnygrpc/grpc.go
+++ b/wrappers/hnygrpc/grpc.go
@@ -18,6 +18,23 @@ import (
 	"google.golang.org/grpc/status"
 )
 
+// This is a map of GRPC request header names whose values will be retrieved
+// and added to handler spans as fields with the corresponding name.
+//
+// Header names must be lowercase as the metadata.MD API will have normalized
+// incoming headers to lower.
+//
+// The field names should turn dashes (-) into underscores (_) to follow
+// precident in HTTP request headers and the patterns established and in
+// naming patterns in OTel attributes for requests.
+var headersToFields = map[string]string{
+	"content-type":      "request.content_type",
+	":authority":        "request.header.authority",
+	"user-agent":        "request.header.user_agent",
+	"x-forwarded-for":   "request.header.x_forwarded_for",
+	"x-forwarded-proto": "request.header.x_forwarded_proto",
+}
+
 // getMetadataStringValue is a simpler helper method that checks the provided
 // metadata for a value associated with the provided key. If the value exists,
 // it is returned. If the value does not exist, an empty string is returned.
@@ -82,13 +99,6 @@ func addFields(ctx context.Context, info *grpc.UnaryServerInfo, handler grpc.Una
 		}
 	}
 
-	headersToFields := map[string]string{
-		"content-type":      "request.content_type",
-		":authority":        "request.header.authority",
-		"user-agent":        "request.header.user_agent",
-		"x-forwarded-for":   "request.header.x_forwarded_for",
-		"x-forwarded-proto": "request.header.x_forwarded_proto",
-	}
 	md, ok := metadata.FromIncomingContext(ctx)
 	if ok {
 		for headerName, fieldName := range headersToFields {
--- a/wrappers/hnygrpc/grpc.go
+++ b/wrappers/hnygrpc/grpc.go
@@ -72,22 +72,19 @@ func addFields(ctx context.Context, info *grpc.UnaryServerInfo, handler grpc.Una
 	span.AddField("handler.name", handlerName)
 	span.AddField("handler.method", info.FullMethod)
 
+	headersToFields := map[string]string{
+		"content-type":      "request.content_type",
+		":authority":        "request.header.authority",
+		"user-agent":        "request.header.user_agent",
+		"x-forwarded-for":   "request.header.x_forwarded_for",
+		"x-forwarded-proto": "request.header.x_forwarded_proto",
+	}
 	md, ok := metadata.FromIncomingContext(ctx)
 	if ok {
-		if val, ok := md["content-type"]; ok {
-			span.AddField("request.content_type", val[0])
-		}
-		if val, ok := md[":authority"]; ok {
-			span.AddField("request.header.authority", val[0])
-		}
-		if val, ok := md["user-agent"]; ok {
-			span.AddField("request.header.user_agent", val[0])
-		}
-		if val, ok := md["x-forwarded-for"]; ok {
-			span.AddField("request.header.x_forwarded_for", val[0])
-		}
-		if val, ok := md["x-forwarded-proto"]; ok {
-			span.AddField("request.header.x_forwarded_proto", val[0])
+		for headerName, fieldName := range headersToFields {
+			if val, ok := md[headerName]; ok {
+				span.AddField(fieldName, val[0])
+			}
 		}
 	}
 }
--- a/wrappers/hnygrpc/grpc.go
+++ b/wrappers/hnygrpc/grpc.go
@@ -2,6 +2,7 @@ package hnygrpc
 
 import (
 	"context"
+	"net"
 	"reflect"
 	"runtime"
 
@@ -13,6 +14,7 @@ import (
 
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/metadata"
+	"google.golang.org/grpc/peer"
 	"google.golang.org/grpc/status"
 )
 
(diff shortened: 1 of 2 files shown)


wrappers/hnygrpc/grpc.go
```diff
@@ -72,16 +74,26 @@ func addFields(ctx context.Context, info *grpc.UnaryServerInfo, handler grpc.Una
...
-	md, ok := metadata.FromIncomingContext(ctx)
+	pr, ok := peer.FromContext(ctx)
 	if ok {
-		if val, ok := md["content-type"]; ok {
-			span.AddField("request.content_type", val[0])
-		}
-		if val, ok := md[":authority"]; ok {
-			span.AddField("request.header.authority", val[0])
+		if pr.Addr != net.Addr(nil) {
+			span.AddField("request.remote_addr", pr.Addr.String())
```
robbkidd, line 80: > But it's a "peer" not a remote.

I opted for naming parity with HTTP requests. `remote_addr` will be the IP and socket of actual HTTP/GRPC remote endpoint talking to this instrumented service. It could be the actual GRPC client or it could be the IP:port of a proxy shimmed in between.

wrappers/hnygrpc/grpc.go
```diff
@@ -72,16 +74,26 @@ func addFields(ctx context.Context, info *grpc.UnaryServerInfo, handler grpc.Una
...
 		}
-		if val, ok := md["user-agent"]; ok {
-			span.AddField("request.header.user_agent", val[0])
+	}
+
+	headersToFields := map[string]string{
+		"content-type":      "request.content_type",
+		":authority":        "request.header.authority",
```
ismith, line 86: Does this want a leading : unlke all the others?

wrappers/hnygrpc/grpc_test.go
```diff
@@ -78,9 +78,11 @@ func TestUnaryInterceptor(t *testing.T) {
 	beeline.Init(beeline.Config{Client: client})
 
 	md := metadata.New(map[string]string{
-		"content-type": "application/grpc",
-		":authority":   "api.honeycomb.io:443",
-		"user-agent":   "testing-is-fun",
+		"content-type":      "application/grpc",
+		":authority":        "api.honeycomb.io:443",
```
ismith, line 82: Same q re above and leading :

ismith (changes requested):

wrappers/hnygrpc/grpc.go
robbkidd: Yup. [RFC 7540 pseudo headers](https://www.rfc-editor.org/rfc/rfc7540#section-8.1.2.3). The colon was new to me, too.

wrappers/hnygrpc/grpc_test.go
robbkidd: Same answer: the leading colon is intentional.

wrappers/hnygrpc/grpc.go
```diff
@@ -72,16 +74,26 @@ func addFields(ctx context.Context, info *grpc.UnaryServerInfo, handler grpc.Una
...
+		"user-agent":        "request.header.user_agent",
+		"x-forwarded-for":   "request.header.x_forwarded_for",
+		"x-forwarded-proto": "request.header.x_forwarded_proto",
+	}
+	md, ok := metadata.FromIncomingContext(ctx)
+	if ok {
+		for headerName, fieldName := range headersToFields {
```
robbkidd, line 93: I'm rethinking this loop. Nicer (I think) to read, but is a few new memory allocations in instrumentation hot path.

MikeGoldsmith: I think this loop is good 👍🏻

wrappers/hnygrpc/grpc.go
```diff
@@ -72,16 +74,26 @@ func addFields(ctx context.Context, info *grpc.UnaryServerInfo, handler grpc.Una
...
-		if val, ok := md[":authority"]; ok {
-			span.AddField("request.header.authority", val[0])
+		if pr.Addr != net.Addr(nil) {
+			span.AddField("request.remote_addr", pr.Addr.String())
 		}
-		if val, ok := md["user-agent"]; ok {
-			span.AddField("request.header.user_agent", val[0])
+	}
+
+	headersToFields := map[string]string{
```
MikeGoldsmith, line 84: This should be made a const so it doesn't get redeclared for each call to `addFields`.

MikeGoldsmith (approved):
Good work @robbkidd 👍🏻 

I've left one suggestion to make the `headersToFields` map a const.

wrappers/hnygrpc/grpc.go
robbkidd: I would love to, but Go doesn't support complex types like maps and slices to be declared directly as constants. 

Options that if you squint kind of look like a constant:

* a function -  We could have a private function that returns a map, but my understanding is that this would still be relentlessly allocating new memory for the local variable.

```go
func getHeadersToFields() map[string]string {
	return map[string]string{
		"content-type":      "request.content_type",
		":authority":        "request.header.authority",
		"user-agent":        "request.header.user_agent",
		"x-forwarded-for":   "request.header.x_forwarded_for",
		"x-forwarded-proto": "request.header.x_forwarded_proto",
	}
}
```

* a package var - We could declare it as a variable in this package and be diligent about not mutating it.

```go
package hnygrpc

import (
	… 
)

var headersToFields = map[string]string{
	"content-type":      "request.content_type",
	":authority":        "request.header.authority",
	"user-agent":        "request.header.user_agent",
	"x-forwarded-for":   "request.header.x_forwarded_for",
	"x-forwarded-proto": "request.header.x_forwarded_proto",
}
```

MikeGoldsmith: Ah, I'm sure I've suggested making a map a const before too 😄 

I like the var option, with comment to say do not update 👍🏻

wrappers/hnygrpc/grpc.go
```diff
@@ -72,16 +74,26 @@ func addFields(ctx context.Context, info *grpc.UnaryServerInfo, handler grpc.Una
...
 
-	md, ok := metadata.FromIncomingContext(ctx)
+	pr, ok := peer.FromContext(ctx)
 	if ok {
-		if val, ok := md["content-type"]; ok {
-			span.AddField("request.content_type", val[0])
-		}
-		if val, ok := md[":authority"]; ok {
-			span.AddField("request.header.authority", val[0])
+		if pr.Addr != net.Addr(nil) {
```
kentquirk, line 79: I actually had to do a couple of minutes of poking around in docs to realize that this was just a test that pr.Addr is not nil. (Yay Go!).  (I initially thought that net.Addr(nil) was a function call that would retrieve the local machine address or something.)
```suggestion
		// if we have an address, put it on the span
		if pr.Addr != net.Addr(nil) {
```

wrappers/hnygrpc/grpc.go
kentquirk: private package var is a fairly common solution to this, yes. 

And just to be pedantic about it, Go doesn't allocate memory on the heap for a locally-defined map that isn't modified, but it does reserve space for it on the stack and has to initialize it. So it costs CPU cycles but doesn't add memory pressure or allocations.

kentquirk (approved):
Just notes. Looks good!

wrappers/hnygrpc/grpc.go
robbkidd: > … just to be pedantic …

I appreciate the details!

The More You Know

wrappers/hnygrpc/grpc.go
robbkidd: We'll ship the loop! 🔁

wrappers/hnygrpc/grpc.go
robbkidd: Local variable moved up to a package variable ([194b95f](https://github.com/honeycombio/beeline-go/pull/341/commits/194b95f72b8817366852076fcc536e9278df5be0)).

wrappers/hnygrpc/grpc.go
ismith: Sweet - thanks for indulging my curiosity.

ismith (approved):
Question answered, fire at will.

PR #345, opened 2022-09-30
Include Commit/Rollback in db.call names

glenjamin:

The `db` instrumentation uses the call stack to identify the operation being recorded, matching them against a list of known operations.

The actual instrumentation code does record calls to `Commit` and `Rollback` - although these are often missing in the trace unless the non-standard `CommitContext` and `RollbackContext` are used (which use the context for trace lineage but otherwise delegate to the non-context Commit/Rollback of the underling db driver)

Currently these operations are not included in the list of know operations, leaving the spans to be named `db` and the `db.caller` field to be `(Commit|Rollback)Context`.

- Adds the relevant commit and rollback operations to the list of known db methods


pkanal (approved):
This looks good to me! 👍🏽

PR #350, opened 2022-10-07
[maint] Drop support for go 1.16

pkanal:
This PR drops support for Go 1.16, it's more than two versions old and we're running into dependency update issues by continuing to support it.

- Remove 1.16 from circleci test matrix
- Update README
- Update go.mod


vreynolds (approved):
🧹

PR #374, opened 2023-03-27
chore: remove projects workflow

vreynolds:


vreynolds:
board have a limit to number of workflows 😭

PR #406, opened 2023-11-28
perf: speed up adding fields, reduce memalloc if field name is already prefixed with "app."

### LINKED ISSUE #401: perf: "app." + key is allocating new objects still

robbkidd:

- Closes #401 

There is a remarkable amount of memory allocation occurring under load to perform the "app." prefixing of field names by the Beeline.

This change skips the memory allocations needed for string concat and usage if the "app." prefix is already present on the field name provided.

### Benchmarks

Existing behavior is no slower or memory hungry, but for every field name provided by the Beeline user that starts with "app.", that is one less memory allocation of the size of the field name string.

```
BenchmarkBeelineAddField/oldAddField/whatever
BenchmarkBeelineAddField/oldAddField/whatever-12                19654003                60.52 ns/op            8 B/op          1 allocs/op
BenchmarkBeelineAddField/AddField/no-prefix
BenchmarkBeelineAddField/AddField/no-prefix-12                  18939754                60.65 ns/op            8 B/op          1 allocs/op
BenchmarkBeelineAddField/AddField/half-prefixed
BenchmarkBeelineAddField/AddField/half-prefixed-12              22405790                51.22 ns/op            4 B/op          0 allocs/op
BenchmarkBeelineAddField/AddField/all-prefixed
BenchmarkBeelineAddField/AddField/all-prefixed-12               27381916                42.88 ns/op            0 B/op          0 allocs/op
```

### DIFF
--- a/beeline.go
+++ b/beeline.go
@@ -274,8 +274,10 @@ func Close() {
 // add.This function is good for span-level data, eg timers or the arguments to
 // a specific function call, etc.
 //
-// Field keys added will be prefixed with `app.` if the key does not already have
-// that prefix.
+// Field keys added will be prefixed with 'app.' if the prefix is not already
+// present on the key name. If you provide a key that starts with 'app.', both
+// speed and memory allocations are improved, especially within hot paths of
+// your application.
 //
 // Errors are treated as a special case for convenience: if `val` is of type
 // `error` then the key is set to the error's message in the span.
@@ -302,8 +304,10 @@ func AddField(ctx context.Context, key string, val interface{}) {
 // context that is better scoped to the request than this specific unit of work,
 // eg user IDs, globally relevant feature flags, errors, etc.
 //
-// Field keys added will be prefixed with `app.` if the key does not already have
-// that prefix.
+// Field keys added will be prefixed with 'app.' if the prefix is not already
+// present on the key name. If you provide a key that starts with 'app.', both
+// speed and memory allocations are improved, especially within hot paths of
+// your application.
 func AddFieldToTrace(ctx context.Context, key string, val interface{}) {
 	tr := trace.GetTraceFromContext(ctx)
 	if tr != nil {
--- a/beeline.go
+++ b/beeline.go
@@ -272,7 +272,10 @@ func Close() {
 // a Handler, feel free to call AddField freely within your code. Pass it the
 // context from the request (`r.Context()`) and the key and value you wish to
 // add.This function is good for span-level data, eg timers or the arguments to
-// a specific function call, etc. Fields added here are prefixed with `app.`
+// a specific function call, etc.
+//
+// Field keys added will be prefixed with `app.` if the key does not already have
+// that prefix.
 //
 // Errors are treated as a special case for convenience: if `val` is of type
 // `error` then the key is set to the error's message in the span.
@@ -280,13 +283,12 @@ func AddField(ctx context.Context, key string, val interface{}) {
 	span := trace.GetSpanFromContext(ctx)
 	if span != nil {
 		if val != nil {
-			namespacedKey := "app." + key // Avoid excess parsing/allocation work
 			if valErr, ok := val.(error); ok {
 				// treat errors specially because it's a pain to have to
 				// remember to stringify them
-				span.AddField(namespacedKey, valErr.Error())
+				span.AddField(getNamespacedKey(key), valErr.Error())
 			} else {
-				span.AddField(namespacedKey, val)
+				span.AddField(getNamespacedKey(key), val)
 			}
 		}
 	}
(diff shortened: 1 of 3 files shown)


beeline_test.go
```diff
@@ -105,8 +106,46 @@ func BenchmarkBeelineAddField(b *testing.B) {
...
+		}
+	})
+	b.Run("AddField/all-prefixed", func(b *testing.B) {
+		for n := 0; n < b.N; n++ {
+			AddField(ctx, "app.foo", 1)
+		}
+	})
+}
+
+func oldAddField(ctx context.Context, key string, val interface{}) {
```
robbkidd, line 136: I intend to remove this function and its benchmark after the PR is reviewed and before merge.

kentquirk (approved):
This is a good call, and provides a nice on-ramp to adding the app prefix explicitly.

beeline.go
```diff
@@ -272,21 +272,23 @@ func Close() {
...
 	if span != nil {
 		if val != nil {
-			namespacedKey := "app." + key // Avoid excess parsing/allocation work
 			if valErr, ok := val.(error); ok {
 				// treat errors specially because it's a pain to have to
 				// remember to stringify them
-				span.AddField(namespacedKey, valErr.Error())
+				span.AddField(getNamespacedKey(key), valErr.Error())
 			} else {
-				span.AddField(namespacedKey, val)
```
MikeGoldsmith, line 289: Why move the call to `getNamespacedKey` into both branches? I think it makes it harder to read when a function call is used as a parameter, and both branches use the the value. I don't think this will reduce allocations, because the key will be resolved before passing into the next func call.

MikeGoldsmith (approved):
I think this is the right way to go. The work done in #402 by introducing a cache introduced too much lock contention. This will allow users of the beeline to improve perf. by doing the `app.` prefix themselves.

The query is not blocking and I don't think it'll have a material impact.

beeline.go
robbkidd: > Why move the call to getNamespacedKey into both branches?

Because I have Ruby in my blood and we call methods anywhere we please! 

I was thinking "don't allocate for a local variable to use something only once—either one path or the other—for the outer AddField() function invocation." You are right, though, there is no additional allocation (work shown below) and therefore I'm all in favor of making things easier to read.

<details><summary>🖱️👈🏻 👀  Show The Work</summary>

```diff
diff --git a/beeline.go b/beeline.go
index 9ce3dbb..2c9a59e 100644
--- a/beeline.go
+++ b/beeline.go
@@ -283,12 +283,13 @@ func AddField(ctx context.Context, key string, val interface{}) {
        span := trace.GetSpanFromContext(ctx)
        if span != nil {
                if val != nil {
+                       namespacedKey := getNamespacedKey(key)
                        if valErr, ok := val.(error); ok {
                                // treat errors specially because it's a pain to have to
                                // remember to stringify them
-                               span.AddField(getNamespacedKey(key), valErr.Error())
+                               span.AddField(namespacedKey, valErr.Error())
                        } else {
-                               span.AddField(getNamespacedKey(key), val)
+                               span.AddField(namespacedKey, val)
                        }
                }
        }
```

… and … 

```
❯ go test -benchmem -run=^$ -bench ^BenchmarkBeelineAddField$ github.com/honeycombio/beeline-go -v
goos: darwin
goarch: arm64
pkg: github.com/honeycombio/beeline-go
BenchmarkBeelineAddField
BenchmarkBeelineAddField/oldAddField/whatever
BenchmarkBeelineAddField/oldAddField/whatever-12                19292650                62.51 ns/op            8 B/op          1 allocs/op
BenchmarkBeelineAddField/AddField/no-prefix
BenchmarkBeelineAddField/AddField/no-prefix-12                  18867331                62.64 ns/op            8 B/op          1 allocs/op
BenchmarkBeelineAddField/AddField/half-prefixed
BenchmarkBeelineAddField/AddField/half-prefixed-12              22028452                52.60 ns/op            4 B/op          0 allocs/op
BenchmarkBeelineAddField/AddField/all-prefixed
BenchmarkBeelineAddField/AddField/all-prefixed-12               26803638                44.19 ns/op            0 B/op          0 allocs/op
PASS
ok      github.com/honeycombio/beeline-go       6.127s
```

… so yea, as you've said before, there's no difference in allocations between assigning a local variable in AddField() and having a function return a value that's immediately used as a parameter still within the scope of AddField().

</details>

PR #416, opened 2024-01-05
feat: adds overridable trace fields

cursedquail:
Adds the concept of an overridable field to traces - this is similar to trace-level fields, but it can be overriden by spans in the same process

The "same process" bit was my attempt at finding a compromise between "propagated fields aren't overridable and I can't change that" and "these fields should probably be propagated".

other things in that design space that I considered:
- not propagating them (idk how much ppl rely on propagation but I didn't love this)
- encoding it into the headers somehow (how??? also we don't control all the formats)

I'm very willing to change how that works, but it's definitely a surprising part of the design space, and while I hope it doesn't doom this change entirely, I'd understand if that ended up being the case :(

Uses the following libhoney-go changes:
- https://github.com/honeycombio/libhoney-go/pull/235

### DIFF
--- a/trace/trace.go
+++ b/trace/trace.go
@@ -124,6 +124,23 @@ type addFieldOptions struct {
 	isOverridable *bool
 }
 
+// Overridable will allow spans created within the same process to override this field with a different value.
+//
+// Overriding a field on a span does not override the entire subtree - in the following hypothetical
+// trace, overriding a field on the parent span will not override that field in the child span.
+//
+//	root
+//	  parent
+//	    child
+//
+// Overridable fields are sent to downstream services, but we cannot encode the "overrideability"
+//
+//	of the field in the propagation context, so the field will cease to be overrideable in the downstream
+//	service.
+//
+// Overridable is best suited to niche cases where you are not propagating the span context,
+//
+//	or cases where you only wish to overwrite the field in a small number of cases, usually in leaf spans.
 func Overridable(overridable bool) addFieldOptions {
 	return addFieldOptions{isOverridable: &overridable}
 }


MikeGoldsmith:
I think having the option to force setitng attribute on a given span and ignore trace level fields is a nice feature. I would go with a simpler API though and use the options design pattern to extend the `addField` function instead of creating new func.

A simplified example could look something like this: https://go.dev/play/p/fm0HYyADUuu

It would the same way with the field and value being added to an intermediate map which is then applied to the span again after trace level fields.

MikeGoldsmith (changes requested):
Would like to see this design use options pattern instead.

trace/trace.go
```diff
@@ -275,6 +330,18 @@ func (s *Span) AddField(key string, val interface{}) {
 	}
 }
 
+func (s *Span) AddFieldIfUnset(key string, val any) {
```
MikeGoldsmith, line 333: I'm unsure what the purpose of this function is? It looks like a direct copy of the `AddField` func above except it uses an `any` instead of `interface{}`. The function name implies it will only set the field if it doesn't already exist.

MikeGoldsmith (changes requested):
Thanks @NLincoln  - the changes to the options pattern look good.

I've left a query regarding one of the new funcs.

MikeGoldsmith:
I think this is moslty there now. I've added a suggestion to use the same options design pattern for the IfNotSet option in libhoney go too. Once that's resolved, we can update this PR to use it and move forward.

Thanks for your patience.

MikeGoldsmith (changes requested):
As above, waiting on some changes to be made in libhoney-go first.

PR #421, opened 2024-02-29
feat: support classic Ingest Keys

jharley:

We've now released Ingest Keys, but in order for them to work with Classic environments properly we need to update the key detection logic.

- Updates the logic of the existing `IsClassicKey` function to support Classic Ingest Keys
- upgrades `libhoney-go` to 1.22.0


beeline.go
```diff
@@ -101,8 +101,7 @@ type Config struct {
 }
 
 func IsClassicKey(config Config) bool {
-	// classic key has 32 characters
-	return len(config.WriteKey) == 32
+	return (&libhoney.Config{APIKey: config.WriteKey}).IsClassic()
```
jharley, line 104: This is .. not beautiful, but it also isn't in the critical performance path.

I figure we have a few options here:

1. leave it as it is
1. duplicate the logic from libhoney anyway
1. go back and make a public function in libhoney that is easier to call in this context?
1. something else I haven't thought of 🙃

cewkrupa: My preference would be for 3, as I think that's closer to the approach we landed on for the other libhoneys where we add a static class method. `libhoney.IsClassicAPIKey(config.WriteKey)` feels like it would be a lot nicer here

jharley: #3 is implemented here if we're happy with it https://github.com/honeycombio/libhoney-go/pull/239

MikeGoldsmith: Approved the libhoney-go alternative func. I guess this func can be removed and just use the new libhoney-go IsClassicKey once it's avaialble.

jharley: I left the local `IsClassicKey` in place as a wrapper around the libhoney function as it is a public function and I didn't want to make an unnecessary breaking change.

MikeGoldsmith (approved):

kentquirk (approved):

PR #455, opened 2025-03-21
feat(trace): update profileIDLabelName to match Pyroscope

lizthegrey:

- Adhere to format Pyroscope OSS expects per https://github.com/grafana/otel-profiling-go/commit/207e85a83187411fc33a161f38b64e90289e6302#diff-e4d4c63738d3e4e1bf4f34e59bf9df3dfc89e9eae421f2345b98dd4535125903

We've switched to using Pyroscope OSS internally rather than Pyroscope/Grafana Cloud, so we need to use the format that the current profiling solution expects for how span names are tagged when doing correlation of span and profile signals.

### DIFF
--- a/trace/context.go
+++ b/trace/context.go
@@ -9,7 +9,7 @@ import (
 const (
 	honeySpanContextKey  = "honeycombSpanContextKey"
 	honeyTraceContextKey = "honeycombTraceContextKey"
-	profileIDLabelName   = "profile_id"
+	profileIDLabelName   = "span_id"
 )
 
 var (


mjayaram (approved):
🎉

kentquirk (approved):
Beelines are being sunset, so try hard not to change them unless we absolutely have to. Can you please add a little more text to the PR summary explaining the context and why this change is needed? This looks like a breaking change but after digging I can see that the spec changed a while back.

lizthegrey:
> Beelines are being sunset, so try hard not to change them unless we absolutely have to. Can you please add a little more text to the PR summary explaining the context and why this change is needed? This looks like a breaking change but after digging I can see that the spec changed a while back.

Sure thing. This is purely a change needed for _internal_ purposes, if this helps.

PR #471, opened 2026-04-13
feat: add bulk AddFields method to Span

lizthegrey:
## Summary

Adds `Span.AddFields(fields map[string]interface{})` which writes all entries to the underlying libhoney Event in a single `eventLock` acquisition, instead of N individual `AddField` calls with a lock acquire/release each.

Depends on honeycombio/libhoney-go#275 (`Event.AddFields`).

We know beeline-go is in maintenance mode, but this is needed for internal CPU savings in shepherd — the per-field beeline `AddField` path accounts for 5.95% cumulative CPU in production profiles while we still have services using beeline-go.

## How to Test

- `go test ./trace/` passes
- `TestAddFields` verifies bulk write on both root and child spans

🤖 Generated with [Claude Code](https://claude.com/claude-code)


trace/trace.go
```diff
@@ -275,6 +275,17 @@ func (s *Span) AddField(key string, val interface{}) {
 	}
 }
 
+// AddFields adds all key/value pairs from the map to this span in a single
+// lock acquisition. More efficient than calling AddField in a loop.
+// Errors in the map are converted to their message string, matching AddField.
```
dstrelau, line 280: Are they? `AddField` has special logic for this.

dstrelau (approved):

trace/trace.go
lizthegrey: done.

PR #477, opened 2026-04-13
feat: change Span.AddFields to accept iter.Seq2[string, any]

lizthegrey:

Companion to honeycombio/libhoney-go#278. The `Span.AddFields(map[string]interface{})` API forces callers to build an intermediate map. Changing to `iter.Seq2[string, any]` lets callers yield key/value pairs directly.

- Change `Span.AddFields` from `map[string]interface{}` to `iter.Seq2[string, any]`
- Error→string conversion done in a wrapping iterator (consistent with `AddField` behavior)
- Tests updated to use `maps.All()`
- Depends on libhoney-go commit c9910b3

Part of a 3-PR chain: libhoney-go → **beeline-go** → hound

🤖 Generated with [Claude Code](https://claude.com/claude-code)


lizthegrey:
Closing for now — going to try the simpler map-based approach first. Branch preserved.

